PR #14, opened 2016-02-12
Update planet-feeds.nix

qknight:
my blog url changed. i would love if you could update it.


qknight:
i had to fix my implementation of the feed-generator and it now contains the article which is what planet.planet actually wants IIRC.

sorry for the confusion.

please merge it now.

PR #19, opened 2016-09-04
nixos-org: add site wide notice for nixos wiki

garbas:
my attempt to create a side wide notice for nixos wiki page

NixOS/nixpkgs#16918

if i correctly understand links bellow this should do the trick:
- https://www.mediawiki.org/wiki/Manual:Interface/Sitenotice
- https://www.mediawiki.org/wiki/Manual:$wgSiteNotice

### DIFF
--- a/nixos-org/webserver.nix
+++ b/nixos-org/webserver.nix
@@ -154,6 +154,8 @@ in
                 #skins = [ ./wiki-skins ];
                 extraConfig =
                   ''
+                    $wgSiteNotice = "The NixOS wiki is outdated and is being shut down. If you wish to add or improve NixOS documentation, please grab a ticket from the [https://github.com/NixOS/nixpkgs/milestone/8 Move the wiki!] milestone.";
+
                     #$wgEmailConfirmToEdit = true;
 
                     #$wgDebugLogFile = "/tmp/mediawiki_debug_log.txt";


nixos-org/webserver.nix
```diff
@@ -154,6 +154,8 @@ in
                 #skins = [ ./wiki-skins ];
                 extraConfig =
                   ''
+                    $wgSiteNotice = "The NixOS wiki is outdated and is being shut down. If you wish to add or improve NixOS documentation, please grab a ticket from the [https://github.com/NixOS/nixpkgs/milestone/8 Move the wiki!] milestone.";
```
domenkozar: s/ticket/issue/.

I think we should also say that each wiki page has an issue, so that people will know how to proceed if they want to add something based on the current page.

garbas: 1/4 of the issues were already closed. i would just point to the milestone, since all that matters is what still needs to be migrated.

i do think that the message can be improved. i never really trust myself to explain stuff correctly :)

simoncarpinter: Speaking as someone caught in the middle of this: I'd suggest pointing to an issue which clearly lays out why the wiki is moving, where it's moving to, how contributors can help out and then links to the milestone work queue.

Essentially: the message you've written here and the title of the milestone don't provide enough information for contributors to help out :).  It's confusing to read "move the wiki!" -- but then have no idea how to proceed.

Many of the issues in the milestone only link to a page to be moved.  Digging through issues to discover the process one should follow is frustrating.

ericsagnes:
Can this be merged any soon?
It seems that many users lose time by trying to follow wiki informations, and these are getting outdated enough so the wiki start doing more harm than good.

garbas:
@rbvermaa could we merge this?

nixos-org/webserver.nix
garbas: @simoncarpinter while i do agree with you, i don't see how we can point directly to github issue. if you have any idea how to do this i'm all on improving this.

maybe a next step would be to add a description to miletstone which would explain more.

edolstra:
Thanks, applied!

PR #28, opened 2017-03-30
NixOS 17.03

globin:
As soon as https://hydra.nixos.org/build/51054560 passes we have a stable 17.03 channel bump.

cc @edolstra


domenkozar:
It passed /cc @rbvermaa @edolstra

PR #30, opened 2017-05-03
Remove Dead and Deprecated Web Services

grahamc:
- repoman 404's so remove the svn redirect
 - remove the long-commented svn subservice
 - stop serving the wiki, remove related redirects, and remove related pgsql config

I don't mean this to be a forever thing, I think having the wiki in a middle phase between "is old and out of date and not fixable" and "we're wanting to fix it soon(?)" is the worst option. Nobody sees the "being shut down" message.

cc @copumpkin

### DIFF
--- a/nixos-org/webserver.nix
+++ b/nixos-org/webserver.nix
@@ -34,13 +34,13 @@ let
           User-agent: *
           Disallow: /repos/
           Disallow: /irc/
-          Disallow: /wiki/
         '';
 
       extraConfig =
         ''
           MaxKeepAliveRequests 0
 
+          Redirect /wiki https://nixos.org/nixos/wiki.html
           Redirect /binary-cache http://cache.nixos.org
           Redirect /releases/channels /channels
           Redirect /tarballs http://tarballs.nixos.org
@@ -119,13 +119,6 @@ in
         }
 
         (nixosVHostConfig // {
-          extraConfig = nixosVHostConfig.extraConfig +
-            ''
-              Redirect /wiki https://nixos.org/wiki
-            '';
-        })
-
-        (nixosVHostConfig // {
           enableSSL = true;
           sslServerKey = "${acmeKeyDir}/key.pem";
           sslServerCert = "${acmeKeyDir}/fullchain.pem";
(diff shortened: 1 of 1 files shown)


nixos-org/webserver.nix
```diff
@@ -308,13 +221,8 @@ in
       synchronous_commit = off
     '';
     authentication = mkOverride 10 ''
-      local mediawiki all ident map=mwusers
       local all       all ident
     '';
-    identMap = ''
-      mwusers root   mediawiki
-      mwusers wwwrun mediawiki
-    '';
```
edolstra: Since Postgres is only used for the wiki, it can be removed entirely.

grahamc: 👍 done

copumpkin:
Rather than take down the wiki webserver entirely, can you just replace it with a landing page saying something like:

> The NixOS wiki is currently down. Please check <manual URLs> <@domenkozar's cookbooks> for documentation in the meantime, and follow https://github.com/NixOS/nixpkgs/issues/22599 for updates on the wiki itself.

It seems friendlier than just making it disappear when it's by far the greatest target of Google links for most Nix searches.

grahamc:
I didn't expect you to have such a positive response! I thought for sure you'd just ask for another week or something :)

copumpkin:
I agree with you that we're in the worst possible situation right now, and I have very little time for anything much, so I don't want to leave things shitty due to my lack of time.

copumpkin:
@edolstra can you just keep a backup of the wiki data lying around for a while? Not saying it needs to be publicly accessible, but I'd hate to lose the info forever.

grahamc:
I'm looking in to making an export, there is https://nixos.org/wiki/Special:Export but it is a bit annoying since it is per-category.

grahamc:
Ok, I copied all the pages from https://nixos.org/wiki/Special:AllPages and pasted them in to https://nixos.org/wiki/Special:Export and made an export of all pages with all versions, available for a limited time here: http://next.gsc.io/Nix+Wiki-20170503130016.xml

@edolstra can you mirror this export to somewhere on nixos.org, so it can be linked to from the "down" page?

grahamc:
This PR now has a redirect and a second PR adds the landing page: https://github.com/NixOS/nixos-homepage/pull/138

I think this and the other are ready.

grahamc:
Rebased on master after the other PR merged. Ready to be merged.

PR #32, opened 2017-07-25
Enable HSTS and fixup http:// links

grahamc:
HSTS tells the browser that you should always use HTTPS when
talking to the domain, for the next max-age. This means it will
(in-browser) redirect from http to https, without hitting a
the webserver.

 - max-age is set to 1 day in case of problems, in the future I'd
   like to extend this.
 - we can't enable includeSubdomains because the following
   domains don't use TLS:

   - www.nixos.org (redirects to nixos.org)
   - releases.nixos.org (deprecated? how much traffic does this get?)
   - planet.nixos.org

  If we can fix these, we can add includeSubdomains too.

Obviously I haven't tested this, as I don't have a local copy of
nixos.org running :)

### DIFF
--- a/delft/delft-webserver.nix
+++ b/delft/delft-webserver.nix
@@ -73,7 +73,7 @@ in
 
         { # Catch-all site.
           hostName = "old.nixos.org";
-          globalRedirect = "http://nixos.org/";
+          globalRedirect = "https://nixos.org/";
         }
 
         { hostName = "buildfarm.st.ewi.tudelft.nl";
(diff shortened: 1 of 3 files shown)


grahamc:
Closes https://github.com/NixOS/nixos-org-configurations/issues/26

grahamc:
Fixed conflicts

edolstra:
Thanks, seems to work!

grahamc:
What a joy! In a few days I'll open a PR to extend max-age to a month or longer. Thank you!

PR #38, opened 2017-09-18
www.nixos.org: serve an SSL certificate

grahamc:
Currently https://www.nixos.org/ is an SSL certificate CN mismatch error page. :(


edolstra:
Maybe we should just remove the DNS entry for `www.nixos.org` instead?

grahamc:
I think that will hurt us more than fixing the redirect. Many people (even people like us) still expect a `www` unfortunately.

PR #39, opened 2018-03-05
channels: add 18.03 versions

vcunat:
I assume we can call the channels with the usual names _already_, and they'll be simply considered beta until officially announced to be released in a few weeks.  They're forked -unstable ATM, so nothing dangerous, really.

### DIFF
--- a/nixos-org/hydra-mirror.nix
+++ b/nixos-org/hydra-mirror.nix
@@ -42,6 +42,8 @@ in
 
   systemd =
     fold recursiveUpdate {} [
+      (makeUpdateChannel "nixos-18.03" "nixos/release-18.03/tested")
+      (makeUpdateChannel "nixos-18.03-small" "nixos/release-18.03-small/tested")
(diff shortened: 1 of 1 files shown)


vcunat:
TODO:
 - [ ] probably drop 16.09 too;
 - [ ] aarch64: either manage to squeeze it into regular channel or add it as well.

grahamc:
aarch64 should probably not get its own channel

vcunat:
See around https://github.com/NixOS/nixpkgs/issues/33745#issuecomment-371154834

vcunat:
I wonder why not even `18.03-small` has appeared so far.  It looks OK on Hydra.

vcunat:
@edolstra: any idea why the channels haven't appeared?

edolstra:
I thought I deployed this before, but apparently not.

vcunat:
Great, they seem OK now.

PR #40, opened 2018-03-21
Automate deploying updates to the macs

grahamc:
- Upgrade to Nix 2 across the board
 - Set up more stable garbage collection
 - Fix mac9
 - Use LnL's nix-darwin to manage the version
   of Nix deployed

To update and deploy:

    $ ./update-channels
    $ ./deploy.sh

You MUST already have setup your SSH config to know
how to connect to mac{1..9} but I'm not going to document that
here.


macs/deploy.sh
```diff
@@ -0,0 +1,40 @@
...
+
+for macId in `seq 1 9`; do
+    host="mac$macId"
+    SSHOPTS="$host"
+
+    if ! ssh $SSHOPTS test -f "$drv"; then
+        echo "Copying derivation to $host"
+
+        # A lame version of nix-copy-closure since we can't SSH directly in
+        # as a trusted user
```
edolstra: Why can't we SSH in as a trusted user?

macs/nix/nix-darwin.json
```diff
@@ -0,0 +1,7 @@
+{
```
edolstra: What is this JSON stuff for?

macs/deploy.sh
grahamc: We could! Taking this to PM.

macs/nix/nix-darwin.json
grahamc: To track and pin the versions of the "channels" being deployed, to keep it from being a floating version.

macs/deploy.sh
grahamc: Fixed!

zimbatm:
Is this still relevant @grahamc ?

grahamc:
Nope :)

PR #44, opened 2018-07-10
terraform: introduce DNS resource definitions

zimbatm:
Start mapping all the resources in AWS that can't be done with nixops

### DIFF
--- a/terraform/README.md
+++ b/terraform/README.md
@@ -1,0 +1,22 @@
+# For the bits that are not nixops-able
+
+For now this manages only resources in the main AWS account.
+
+## Usage
+
+Make sure to have the AWS key-pair in the environment, in
+`~/.aws/credentials` or as the EC2 metadata service.
+
+Then run:
+
+```
+nix-shell --run "terraform apply"
+```
+
+## TODO
+
+* Add remote state
+* Add cert manager
+* Manage the S3 buckets and CloudFront distributions
+* Manage the VPCs and Security Groups
+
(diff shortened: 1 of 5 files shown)


rbvermaa:
@zimbatm Why not use nixops for these, as it supports route53 resources?

zimbatm:
It's not able to declare the Route53 zone isn't it? I want to get to a place where we have all the cloud resources declared as code. CloudFront, the S3 buckets, the IAM roles, the security groups and auto-scaling groups, the VPC.

AmineChikhaoui:
@zimbatm isn't it this resource (Route53 zone) https://github.com/NixOS/nixops/blob/master/nix/route53-hosted-zone.nix ?
I guess the advantage of terraform is that it can take over the management of an already provisioned resource while you have to re-provision using nixops.
Btw, S3 buckets/IAM roles/VPC resources are already supported in nixops.

zimbatm:
Yeah I might be mistaken, nixops is awesome and I use it but only for deploying since it has much less manpower available and doesn't support remote state yet. It's easy to hit a corner where some API is not properly mapped and then have a big effort gap. Some day we'll be able to import the terraform plugins into nixops but until then I find it easier to just use terraform.

Since we know that the existing AWS account is going away maybe another question is whenever we want to use it for Route53 or some other DNS provider. Having everything in terraform also helps in terms of data locality, it makes it easy to declare a CloudFront distribution and attach it to a record in one go for example.

zimbatm:
superseded by #68

PR #48, opened 2018-07-11
terraform: describe nixpkgs-tarballs cloudfront

zimbatm:


terraform/README.md
```diff
@@ -0,0 +1,14 @@
...
+
+## Usage
+
+Make sure to have the AWS key-pair in the environment, in
+`~/.aws/credentials` or as the EC2 metadata service.
+
+Then run:
+
+```
+nix-shell --run "terraform apply"
```
grahamc: Maybe add a note about that this will present a diff and ask if you want to apply it, and that `apply` won't just go changing things right away.

zimbatm: Right, I also forgot the `terraform init` bit

terraform/nixpkgs-tarballs.tf
```diff
@@ -0,0 +1,45 @@
+resource "aws_s3_bucket" "nixpkgs-tarballs" {
+  bucket = "nixpkgs-tarballs"
+  region = "eu-west-1"
+  acl    = "public-read"
+
+  website {
+    index_document = "index.html"
+  }
+}
```
edolstra: Note that this bucket is already managed by NixOps.

zimbatm: What happens if it gets removed from NixOps, will NixOps try to destroy it?

terraform/README.md
zimbatm: fixed

terraform/nixpkgs-tarballs.tf
zimbatm: anyways, I just commented out the declaration for now

PR #50, opened 2018-07-13
add packet-epyc-1

grahamc:
Not sure how to actually tell Hydra to use it, though.


grahamc:
(for the archeologists of the future, /etc/nix/machines is modified manually)

PR #61, opened 2018-11-02
Automatic GC if we drop low. Limits debatable...

srhb:
In response to a large build dying from out-of-disk conditions
https://github.com/NixOS/nixpkgs/issues/49442#issuecomment-435320605

### DIFF
--- a/delft/build-machines-common.nix
+++ b/delft/build-machines-common.nix
@@ -13,6 +13,12 @@ with lib;
   # the same time.
   systemd.timers.nix-gc.timerConfig.RandomizedDelaySec = "1800";
 
+  # If we drop below 20GiB during builds, free 20GiB
+  nix.extraOptions = ''
+    min-free = ${toString (20*1024*1024*1024)}
+    max-free = ${toString (40*1024*1024*1024)}
+  '';
+
   users.extraUsers.root.openssh.authorizedKeys.keys = singleton
     ''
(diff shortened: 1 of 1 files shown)


srhb:
@grahamc ping

srhb:
Re. the limits, the question is really how much speed/load saving is won from doing a small gc. If it's slow and expensive anyway (ie. if the constant cost of doing _any_ size GC is high,) we should increase `max-free` a lot. This should probably receive testing on a cross section of the machine types, but this is a starting point.

vcunat:
My feeling from GC on rotating drives is that vast majority of time is taken by unlinking files.

srhb:
@vcunat That sounds like a small GC is indeed preferable. :)

edolstra:
Seems like a good way to give the auto-GC feature some testing :-)

vcunat:
BTW, on `t2a` I've seen running out of inodes while having plenty of free space (more than half).  I plan to work around that just by collecting everything relatively often (once a day or so), since the machine won't benefit much from caching.  Only the `metrics` job is ran there and that's rather special; it seems unlikely to happen on other machines.

zimbatm:
I want to deploy this soon. It requires some careful manual rolling deployment to avoid breaking all of the nodes.

davidak:
Any update on this?

zimbatm:
I haven't touched the NixOS infra in a long time now. Somebody else should probably take over.

PR #80, opened 2019-03-16
channel updates: make sequential

### LINKED ISSUE #79: Channel updates happen all at once, meaning none at all.
Instead, let's consider running them sequentially.  The memory pressure on the instance is too high for it to get two done at once.

grahamc:
Create a sequential ordering with systemd After-relationships between
each channel's update job. Then, remove individual timers and create
one timer for all the jobs.

Fixes #79

### DIFF
--- a/lib/service-order.nix
+++ b/lib/service-order.nix
@@ -1,0 +1,61 @@
+# Ordering Services
+#
+# Given a set of services, make them run one at a time in a specific
+# order, on a timer.
+{ lib }:
+{
+  # Given a list of systemd service, give each one an After
+  # attribute, so they start in a specific order. The returned
+  # list can be converted in to a systemd.services attrset with
+  # `lib.listToAttrs`.
+  #
+  # Example:
+  #
+  #  mkOrderedChain [
+  #    { name = "foo"; value = { script = "true"; }; }
+  #    { name = "bar"; value = { script = "true"; }; }
+  #  ]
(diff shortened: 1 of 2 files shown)


modules/service-order.nix
```diff
@@ -0,0 +1,61 @@
+# Ordering Services
+#
+# Given a set of services, make them run one at a time in a specific
+# order, on a timer.
+{ lib, ... }:
```
samueldr: Any reason to allow additional arguments here?

Since this is an attrset that isn't importable as a module, this doesn't really make sense to me as is. Was this initially started considering it could be implemented as a NixOS module? Could it be made one? Otherwise, would it make make more sense at a location like `../lib` or `./lib`?

grahamc: +1 that was sloppy of me

grahamc: Fixed up.

grahamc:
Deployed!

edolstra:
Wouldn't it be easier to just have a shell script that runs `mirror-branch` in sequence for all supported branches?

grahamc:
I had considered that, but thought it much nicer to be able to start them individually if we have need to, plus have individual logs and result statuses. Would you prefer I replace this with the simpler approach?

edolstra:
Yeah you're right, it's nice to have individual control over the units.

PR #85, opened 2019-04-26
Automatically import Packet.net spot instances to Hydra

grahamc:
Packet lets us get cheap and powerful builders for our Hydra over Spot, but they change regularly. For that reason we don't want to manually update the configuration settings. Here is a simple Python program and systemd timer to import manage them automatically.

I've already deployed this, but would like some eyes on the branch anyway.

### DIFF
--- a/delft/packet-importer.nix
+++ b/delft/packet-importer.nix
@@ -1,0 +1,44 @@
+{ config, lib, pkgs, ... }:
+let
+  importer = pkgs.callPackage ../hydra-packet-importer { };
+in
+{
+  deployment.keys."hydra-packet-import.json" = {
+    keyFile = ../hydra-packet-import.json;
+    user = "hydra-packet";
+  };
+
+  users.extraUsers.hydra-packet = {
+    description = "Hydra Packet Machine Importer";
+    group = "hydra";
+  };
+
+  systemd.tmpfiles.rules = [
+    "d /var/lib/hydra/packet-import 0755 hydra-packet hydra -"
+    "f /var/lib/hydra/packet-import/machines 0644 hydra-packet hydra -"
+  ];
+
+  services.hydra-dev.buildMachinesFiles = [
+    "/var/lib/hydra/packet-import/machines"
+  ];
+
+  systemd.services.hydra-packet-import = {
+    path = with pkgs; [ openssh moreutils ];
+    script = "${importer}/bin/hydra-packet-importer /run/keys/hydra-packet-import.json | sponge /var/lib/hydra/packet-import/machines";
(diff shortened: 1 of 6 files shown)


delft/packet-importer.nix
```diff
@@ -0,0 +1,51 @@
...
+    keyFile = ../hydra-packet-import.json;
+    user = "hydra-packet";
+  };
+
+  users.extraUsers.hydra-packet =
+    { description = "Hydra Packet Machine Importer";
+      group = "hydra";
+    };
+
+  system.activationScripts.hydra-packet = lib.stringAfter [ "users" ]
```
flokli: you should be able to replace the whole activation script by

systemd.tmpfiles.rules = [
  "d /var/lib/hydra/packet-import 0755 hydra-packet hydra -"
  "f /var/lib/hydra/packet-import/machines 0644 hydra-packet hydra -"
];
```

File: delft/packet-importer.nix [hunk 2]
```diff
@@ -0,0 +1,51 @@
...
+    ''
+      mkdir -m 0755 -p /var/lib/hydra/packet-import
+      chown hydra-packet.hydra /var/lib/hydra/packet-import
+      if [ ! -f /var/lib/hydra/packet-import ]; then
+        touch  /var/lib/hydra/packet-import/machines
+        chown hydra-packet.hydra /var/lib/hydra/packet-import/machines
+        chmod 0644 /var/lib/hydra/packet-import/machines
+      fi
+      chown hydra-packet.hydra /var/lib/hydra/packet-import
```
flokli: this is the same as in L19 too.

delft/packet-importer.nix
grahamc: Nice, thanks!

delft/packet-importer.nix
grahamc: Replaced with the tmpfiles rule, right? Pretty cool.

hydra-packet-importer/import.py
```diff
@@ -0,0 +1,100 @@
...
+import sys
+
+def debug(*args, **kwargs):
+    print(*args, file=sys.stderr, **kwargs)
+
+
+def get_devices(manager):
+    devices = []
+
+    page = 'projects/%s/devices?page=%d' % (config['project_id'], 1)
```
domenkozar: A bit more idiomatic python3:

```suggestion
    page = 'projects/{}/devices?page={}'.format(config['project_id'], 1)
```

grahamc: You bet!

PR #89, opened 2019-09-19
ceres: calculate Nix GC's max-freed including inodes

grahamc:
Cere's disk keeps running out of inodes. This increases free disk space to 400g, but for each percentage point over 75%, frees 500mb additional.

### DIFF
--- a/delft/ceres.nix
+++ b/delft/ceres.nix
@@ -50,7 +50,39 @@
   services.hydra-dev.buildMachinesFiles = [ "/etc/nix/machines" ];
 
   nix.gc.automatic = true;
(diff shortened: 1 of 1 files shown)


edolstra:
Looks complicated, isn't it simpler to just set min-free to 400 GB or so?

grahamc:
Might be fine to just raise it to 400, which would be simpler. Of course, it would be nice to actually be able to ask Nix for what we actually want and not dance around this guessing game (same guessing game I do in this PR.). I'll go for the 400M the way it is already in master :)

edolstra:
Yeah, we could add some inode options to the auto-GC in Nix.

PR #93, opened 2020-01-07
status-page: add alerts for github issues

LnL7:
status-page

/cc @grahamc

### DIFF
--- a/delft/eris/status-page/index.html
+++ b/delft/eris/status-page/index.html
@@ -57,6 +57,8 @@
         <h1>NixOS</h1>
       </div>
 
+      <div id="alerts"></div>
+
       <div class="hero12">
         <h2>Nix Channel Status</h2>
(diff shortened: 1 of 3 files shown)


samueldr:
The age labels are not of a specific width, which causes the alert text to be off.

image

Using `display-block` we can set a min-width. Using these it should be correct for most sizes. A 1em margin was kept in case it is larger than 5em.

Additionally, a 10px top margin for the `div#alerts` is required for the appropriate bootstrap 2 metrics. Otherwise it's too close to the navbar.

```diff
diff --git a/delft/eris/status-page/status.css b/delft/eris/status-page/status.css
index 8b3b821..7b6be18 100644
--- a/delft/eris/status-page/status.css
+++ b/delft/eris/status-page/status.css
@@ -12,8 +12,16 @@
   max-width: 15em;
 }
 
+.issue-age {
+	min-width: 6em;
+	display: inline-block;
+}
 .issue-link {
-    margin-left: 3em;
+    margin-left: 2em;
+}
+
+#alerts {
+	margin-top: 10px;
 }
 
 .github {
```

samueldr:
:+1: for the concept, execution is on par with the existing code. Could need a cleanup, but I think it should be handled as a whole rather than on a per-improvement basis.

PR #94, opened 2020-01-07
status-page: fix race condition in initialization.

turboMaCk:
Due to the fact that loading state text is inserted on window onload,
this insertion can potentially happen **after** all other async
actions are already finished and thus replace the actual data.
This fix makes sure that requests are triggered after onload is done.

cc @grahamc

This is what happens when the async actions are resolved with timing that leads to this issue:

image

## Steps to reproduce:

1. make sure that **client side cache is disabled* in browser
    - common way is to set "disable caching while developer console is open" and open the console
1. refresh page several times until you hit critical timing in which issue occures.

### DIFF
--- a/delft/eris/status-page/status.js
+++ b/delft/eris/status-page/status.js
@@ -1,7 +1,12 @@
-window.onload = (_) => {
-    var tbody = document.getElementById("channel-status");
-    tbody.innerHTML = "<tr><td class='jsfallback' colspan='5'>Loading data from Prometheus...</td></tr>";
-};
+function init() {
+  return new Promise(resolve => {
+    window.onload = () => {
+      var tbody = document.getElementById("channel-status");
+      tbody.innerHTML = "<tr><td class='jsfallback' colspan='5'>Loading data from Prometheus...</td></tr>";
+      resolve();
+    };
+  });
+}
 
 function aggregateByChannel(result) {
   return result.reduce((acc, {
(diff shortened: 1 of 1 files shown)


grahamc:
cc @samueldr mind taking a look-see?

delft/eris/status-page/status.js
```diff
@@ -1,7 +1,12 @@
-window.onload = (_) => {
```
turboMaCk: I've removed this wildcard `_` for the argument name because in JS all arguments are optional (with default to undefined) and arrity of functions is pretty much dynamic. While I personally also like this convention of denotating ignored arguments with `_` the problem i JS case is that one can never trust the number of arguments so it can be missleading when someone less experienced is reading the code and making assumptions about arity based on those wildcards even though in this case the arity won't change (unless some monkeypatching gets involved).

to make matters worse there are polupar libraries like [lodash](https://lodash.com/) and [underscore](https://underscorejs.org/) that make use of `_` as a namespace for functions and using `_` as argument name can cause some unfortunete hard to track shadowing of those namespaces.

Not a big deal just wanted but better safe than sorry I would say.

turboMaCk:
indentation of initi changed to 2 space to be consistent with rest of the code. This is ready for merge from my side.

delft/eris/status-page/status.js
```diff
@@ -1,7 +1,13 @@
...
-    var tbody = document.getElementById("channel-status");
-    tbody.innerHTML = "<tr><td class='jsfallback' colspan='5'>Loading data from Prometheus...</td></tr>";
-};
+function init() {
+  return new Promise(resolve => {
+    window.onload = (a, b, c) => {
+      console.log(a,b,c);
```
turboMaCk: leftover :man_facepalming:

PR #99, opened 2020-02-04
delft/eris: drop alerts with severity!='page', ignore stuck channels

tilpner:
As requested, less spam on #nixos-dev. It would be easy to redirect these alerts to e.g. #nixos-infra, but I'll leave that decision to someone else, since I don't know who should respond to ChannelUpdateStuck in what manner and timeframe.


grahamc:
Thank you!

PR #108, opened 2020-03-31
Remove min-free and max-free from darwin configuration

matthewbauer:
This has a bug where gc can happen in the middle of a build:

https://hydra.nixos.org/build/114884087/nixlog/1/tail


zimbatm:
Is this still relevant?

matthewbauer:
No, I think Nix has fixed this bug.

Issue #34, opened 2017-08-17
Clean up domain list


grahamc:
Here is a rough sorting of domains to drop / keep, based on what just redirected back to nixos.org, domains I'd never heard of before, or were showing a "website maintenance" page, plus a couple I have heard of but nobody uses:

Drop:
ts.nixos.org
svn.nixos.org
status.nixos.org * never updated since I came around ...
stan.nixos.org
releases.nixos.org * I wonder how much traffic this gets
mturk.nixos.org
monitor.nixos.org
lucifer.nixos.org
losser.nixos.org
barbrady.nixos.org

Keep:
cache.nixos.org
hydra.nixos.org
weekly.nixos.org
conf.nixos.org
tarballs.nixos.org
planet.nixos.org

Domain list from https://github.com/NixOS/nixos-org-configurations/pull/33#issuecomment-323042489

grahamc:
domain|ready for preload|serves https|redirects to https|serves STS|comments
-|-|-|-|-|-
cache.nixos.org| ✔️ |✔️ | ❌ | ❌ |it'd be ideal to have it also redirect HTTP to HTTPS. I'm not sure this is possible. From what I'm reading, CF can't send STS headers
hydra.nixos.org|✔️ |✔️ |✔️ |✔️ |
weekly.nixos.org|❌ |❌ |❌ |❌ |hosted on github pages, no support for SSL. Move to nixos.org box?
conf.nixos.org |❌ |❌ |❌ |❌ |hosted on github pages, no support for SSL Move to nixos.org box?
tarballs.nixos.org|❌ |❌ |❌ |❌ |hosted on CloudFront. Enable ACM & TLS?`
planet.nixos.org|❌ |❌ |❌ |❌ |hosted on the nixos.org box. enable LetsEncrypt?

zimbatm:
GitHub added support for HTTPS so both conf and weekly are now available over HTTPS

zimbatm:
I'm pretty sure all of that has been cleaned during the Netlify DNS migration.

PR #112, opened 2020-05-23
tf/cache: Redirect HTTP to HTTPS

endocrimes:
Currently cache.nixos.org is available over HTTP and HTTPS. This is subpar for a variety of reasons.

This commit introduces the viewer_protocol_policy of redirect-to-https to ensure that all connections happen over TLS.

Docs: https://www.terraform.io/docs/providers/aws/r/cloudfront_distribution.html#viewer_protocol_policy

Related to: #34

### DIFF
--- a/terraform/cache.tf
+++ b/terraform/cache.tf
@@ -94,7 +94,7 @@ resource "aws_cloudfront_distribution" "cache" {
     allowed_methods        = ["HEAD", "GET"]
     cached_methods         = ["HEAD", "GET"]
     target_origin_id       = "S3-nix-cache"
-    viewer_protocol_policy = "allow-all"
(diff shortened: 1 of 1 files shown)


edolstra:
I don't really see a reason to ban unencrypted access. The binary cache is signed either way. Also http is lower-latency than https (e.g. `curl http://cache.nixos.org/nix-cache-info` takes 0.02s vs 0.05s for https).

endocrimes:
@edolstra that's fair - I mostly filled this when looking through and seeing @grahamc's issue (although HTTPS has advantages beyond just authenticating the packages)

davidak:
So can this be closed?

PR #120, opened 2020-09-04
20.09 channels

worldofpeace:
Ref to past PR https://github.com/NixOS/nixos-org-configurations/pull/100

### DIFF
--- a/channels.nix
+++ b/channels.nix
@@ -47,15 +47,15 @@ rec {
 
       "nixos-19.09" = {
         job = "nixos/release-19.09/tested";
-        current = true;
+        current = false;
       };
       "nixos-19.09-small" = {
         job = "nixos/release-19.09-small/tested";
-        current = true;
+        current = false;
       };
       "nixpkgs-19.09-darwin" = {
         job = "nixpkgs/nixpkgs-19.09-darwin/darwin-tested";
-        current = true;
+        current = false;
       };
   };
 
--- a/channels.nix
+++ b/channels.nix
@@ -19,6 +19,19 @@ rec {
         current = true;
       };
 
+      "nixos-20.09" = {
+        job = "nixos/release-20.09/tested";
+        current = true;
+      };
+      "nixos-20.09-small" = {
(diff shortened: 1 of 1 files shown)


cole-h:
Should 19.09 also be marked EOL here?
https://github.com/NixOS/nixos-org-configurations/blob/a0ebad908a413bbd2f7db2d4b90d611b86a37abb/channels.nix#L5-L6

worldofpeace:
@cole-h It sounds like exactly as you've said it, `19.09` should have been marked EOL here and 19.09 should be removed.
If that's not true someone better correct us :grin:

cole-h:
cc @grahamc as the author and @edolstra as the merger of those instructions: is this indeed what you had in mind?

If so, maybe the comments should be updated (in another PR) to be more actionable/explicit ("when adding a new version, remove the version from 3 releases ago (e.g. for 19.09, remove 18.03) and mark the oldest remaining version as not-current" or something).

AmineChikhaoui:
For reference, I created the jobsets following the previous PR:
https://hydra.nixos.org/jobset/nixos/release-20.09
https://hydra.nixos.org/jobset/nixos/release-20.09-aarch64
https://hydra.nixos.org/jobset/nixos/release-20.09-small
https://hydra.nixos.org/jobset/nixpkgs/staging-20.09
https://hydra.nixos.org/jobset/nixpkgs/nixpkgs-20.09-darwin

vcunat:
status.nixos.org does not seem to be affected by this change so far.  Some deployment step is missing?

cole-h:
I asked this same question on IRC a bit ago -- @AmineChikhaoui is taking care of it (https://logs.nix.samueldr.com/nixos-dev/2020-09-08#1599586187-1599588570).

Judging by the layout of this repo, s.no.o is hosted on delft / eris, while only bastion was redeployed, I believe.

vcunat:
It is there now.  I'm puzzled that 19.09 didn't get visually marked as EOL even though I see it in this diff, but that's such a small thing...

PR #121, opened 2020-09-09
Update planet-feeds.nix

layus:
Add @layus blog to the nixos planet.

Only the articles tagged 'nix' should appear in the added feed.


layus:
Thanks !

PR #131, opened 2020-11-20
status-page: fix navbar dropdown

jtojnar:
Looks like it tries to find element using # selector based on the href attribute:

https://github.com/twbs/bootstrap/blob/v3.4.1/js/dropdown.js#L27

resulting in the following error:

    Uncaught Error: Syntax error, unrecognized expression: #

Explicitly specifying data-target fixes that.

cc @samueldr


garbas:
the status page moved here and i migrated this PR here https://github.com/NixOS/nixos-status/pull/5

PR #136, opened 2020-11-29
20.03 is EOL

jonringer:
related: https://github.com/NixOS/nixpkgs/issues/101975

### DIFF
--- a/channels.nix
+++ b/channels.nix
@@ -34,15 +34,15 @@ rec {
 
       "nixos-20.03" = {
         job = "nixos/release-20.03/tested";
-        current = true;
+        current = false;
       };
       "nixos-20.03-small" = {
         job = "nixos/release-20.03-small/tested";
-        current = true;
+        current = false;
       };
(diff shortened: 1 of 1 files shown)


SuperSandro2000 (approved):

cole-h:
Another question that I think should be answered: is this when we want to drop 19.09 from the status page altogether?

jonringer:
> Another question that I think should be answered: is this when we want to drop 19.09 from the status page altogether?

I'm not confident in what is consuming this information, or in what ways. If someone from infra says yes, then I'll be happy to remove the 19.09 entries altogether.

vcunat:
Still not shown on https://status.nixos.org/  Some re-deploy step is missing?

jonringer:
I'm assuming the homepage needs to update the flakes

cc @garbas

garbas:
@edolstra when you merged this did you deployed to eris? channels.nix are being served by prometheus which is on eris (i think).

PR #137, opened 2020-12-03
Removing status.nixos.org (only static html/css/js)

garbas:
because it is being moved to a separate repository (https://github.com/nixos/nixos-status) to be deployed to Netlify, like we are already doing for https://nixos.org and https://search.nixos.org.

# Why

The reason for this change was that we want to add similar design to https://status.nixos.org as we have it on https://nixos.org. For that it would be nice to have similar workflow as we already have for https://github.com/nixos/nixos-homepage and https://github.com/nixos/nixos-search. And who knows maybe that would allow us to improve https://status.nixos.org page even further.

As we did with https://nixos.org the point of moving is not to move it to Netlify, but to make https://status.nixos.org a static website that we can deploy anywhere. The reasons why Netlify was selected was already discussed when we had this discussion regarding moving https://nixos.org.

# How

And as it was the problem with https://nixos.org the migration is not as straightforward as we would like. We will have to move this in few steps:

- [x] 1. Point monitoring.nixos.org to eris's IP (like we currently do for status.nixos.org)

- [x] 2. Serve https://status.nixos.org/prometheus and https://status.nixos.org/grafana under subdomain alongside existing domain. The proposed subdomain: https://monitoring.nixos.org/prometheus and https://monitoring.nixos.org/grafana. #138 

- [x] 3. Move status html/js/css to a static page and deploy to Netlify. We need to create proxy for `/grafana` and `/prometheus` which point to subdomain. https://github.com/NixOS/nixos-status/pull/3

- [x] 4. Change DNS for status.nixos.org to point at Netlify.

- [ ] 5. Remove status html/css/js and remove serving status.nixos.org with erin. This PR, #137.

- [ ] 6. Move any issues regarding status page to https://github.com/nixos/nixos-status.

Have I missed something?


grahamc:
I'd prefer keep grafana and prometheus on the same domain, with a path. This is in case more tools get put up or something to that effect.

garbas:
@grahamc why would be a bad idea having each tool per subdomain? We don't know what the future holds and we don't want to paint our-self again in the corner like we did with nixos.org and status.nixos.org. I don't think it costs us anything.

On the other side I'm fine with what ever is decided. I thought I would bring this up.

garbas:
Would my "4 step" plan create any (significant) downtime?
Is something else missing from the list?

grahamc:
It is not trivial to add new records, and I don't want to be in a position where we're working around that by adding a subpath to grafana.nixos.org or something weird. Please use a single subdomain. Other than that, I think this will work okay.

grahamc (changes requested):
It is not trivial to add new records, and I don't want to be in a position where we're working around that by adding a subpath to grafana.nixos.org or something weird. Please use a single subdomain. Other than that, I think this will work okay.

garbas:
@grahamc Maybe you didn't know but @edolstra moved DNS servers to Netlify as well and it is a single click to add more. Even sharing the access to this is now possible.

But yeah, if you still like this is not good, i'll use monitoring.nixos.org domain. Can you be more specific where about prometheus and grafana URL?

grahamc:
Ah, nice. However, I would still prefer a single domain for now. Let's do monitoring.nixos.org/prometheus, monitoring.nixos.org/grafana and the root page can be trivial redirect to status.nixos.org where we can maintain the index.

garbas:
PR updated to use `monitoring.nixos.org`.

delft/eris.nix
```diff
@@ -40,20 +40,19 @@ in {
 
   services.nginx = {
     enable = true;
-    virtualHosts."status.nixos.org" = {
+    virtualHosts."monitoring.nixos.org" = {
       enableACME = true;
       forceSSL = true;
-      root = ./eris/status-page;
```
grahamc, line 46: Looks good, but I think nginx will want us to define a root or a / location. I wonder what it takes to make the / a redirect?

garbas: Oh right, I forgot the redirect.

garbas:
Update:
- step 1: `monitoring.nixos.org` now pointing to `138.201.32.77` (erin).
- step 2: https://github.com/NixOS/nixos-org-configurations/pull/138
- step 3: https://github.com/NixOS/nixos-status/pull/3

garbas:
merged master since #138 was merged and deployed.

garbas:
I moved further and redirect status.nixos.org domain at netlify. Now all that needs to be done is a clean up.

garbas:
@edolstra this PR is now ready to be merged and eris redeployed.

garbas:
I tranfered all the issues/prs to nixos-status repo. I consider this migration done.

PR #138, opened 2020-12-03
Serve grafana and prometheus under monitoring.nixos.org

garbas:
This is step 2 from the https://github.com/NixOS/nixos-org-configurations/pull/137


garbas:
@grahamc ping

PR #171, opened 2021-09-07
Bastion terraform

zimbatm:
This moves all the AWS resource declarations to terraform.

The terraform state has been imported from the existing resources and is clean. It's not the cleanest terraform since it's just a straight import.

The nixops state can probably be destroyed and re-created since we don't care about the old resources anymore and the state is not needed anymore.

Now it's possible to `nixos-rebuild switch --flake .#bastion --use-remote-sudo` to deploy from the bastion host.

Fixes #171


edolstra:
I'd prefer not to do this until we can do the entire deployment using *only* terraform. This change will make deployment harder/uglier since you now need two tools to manage the infrastructure. And NixOps cannot reference output values from terraform, so you get ugly things like `deployment.targetHost = "bastion.nixos.org"`.

zimbatm:
Ok. 

Note that the current NixOps deployment is broken and is blocking the fixes to the channel script in https://github.com/NixOS/nixos-channel-scripts/pull/49

The latest commit deploys with terraform, but I'm hitting another issue now: https://github.com/tweag/terraform-nixos/issues/60

grahamc:
> I'd prefer not to do this until we can do the entire deployment using only terraform. This change will make deployment harder/uglier since you now need two tools to manage the infrastructure. And NixOps cannot reference output values from terraform, so you get ugly things like deployment.targetHost = "bastion.nixos.org".

I am not very excited about terraform as a NixOS deploy tool. I've found it a bit slow, and doesn't handle partial network deployments very well. I've had good success using a combination of Terraform + Morph. Perhaps I could demo how that works?

zimbatm:
I don't really care, just pick something so I can deploy the channel script updates. With the implementation in this branch, it will also be possible to `nixos-rebuild switch` from the bastion directly.

zimbatm:
I had to fork tweag/terraform-nixos to make it work with our sub-folder flake.

zimbatm:
Assuming that this is fine.

PR #173, opened 2021-09-10
eris: fix BuildsStuckOverTwoDays alert

lukegb:
This should consider *all* builds stuck over two days, not just those stuck between two days and three days.


lukegb:
cc @grahamc

PR #187, opened 2021-11-22
Include 21.11 channels

tomberek:
@grahamc

PR #189, opened 2021-11-23
Add `variant`, `status` fields to channels.nix, export them in channel-exporter

cole-h:
This can then be exported by prometheus in order to find the most recent
stable channel, as well as filter out variants such as small, aarch64,
and darwin channels.

---

Example query returning the latest stable NixOS version:

    channel_revision{current="1",stable="1",variant=""}

The variant is empty when there is no set variant because `nixos-21.05`
is not a variant but the "base", while `nixos-21.05-small` is a "small"
variant of `nixos-21.05`.

### DIFF
--- a/channels.nix
+++ b/channels.nix
@@ -1,87 +1,87 @@
 rec {
   channels = {
-      # "Channel name" = {
-      #   job = "project/jobset/jobname"; like https://hydra.nixos.org/job/<value>/latest-finished
-      #   # when adding a new version, mark the oldest as "not current". New releases should always be current.
-      #   # Channels where `current = false` are marked `end-of-life` on https://status.nixos.org/ and alerting is disabled.
-      #   current = true;
-      # };
-      "nixos-unstable" = {
-        job = "nixos/trunk-combined/tested";
-        current = true;
-      };
-      "nixos-unstable-small" = {
-        job = "nixos/unstable-small/tested";
-        current = true;
-      };
-      "nixpkgs-unstable" = {
-        job = "nixpkgs/trunk/unstable";
-        current = true;
-      };
-      
-      "nixos-21.11" = {
-        job = "nixos/release-21.11/tested";
-        current = true;
-      };
-      "nixos-21.11-small" = {
-        job = "nixos/release-21.11-small/tested";
-        current = true;
-      };
-      "nixpkgs-21.11-darwin" = {
-        job = "nixpkgs/nixpkgs-21.11-darwin/darwin-tested";
-        current = true;
-      };
-      "nixos-21.11-aarch64" = {
-        job = "nixos/release-21.11-aarch64/tested";
-        current = true;
-      };
+    # "Channel name" = {
+    #   job = "project/jobset/jobname"; like https://hydra.nixos.org/job/<value>/latest-finished
+    #   # when adding a new version, mark the oldest as "not current". New releases should always be current.
+    #   # Channels where `current = false` are marked `end-of-life` on https://status.nixos.org/ and alerting is disabled.
+    #   current = true;
+    # };
+    "nixos-unstable" = {
+      job = "nixos/trunk-combined/tested";
+      current = true;
+    };
+    "nixos-unstable-small" = {
+      job = "nixos/unstable-small/tested";
+      current = true;
+    };
+    "nixpkgs-unstable" = {
+      job = "nixpkgs/trunk/unstable";
+      current = true;
+    };
 
-      "nixos-21.05" = {
-        job = "nixos/release-21.05/tested";
-        current = true;
-      };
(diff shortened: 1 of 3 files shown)


grahamc:
Looks pretty good to me, let's make the `int(about['current'])` code broken out in to the variable definition block and use it elsewhere, similar to the `stable` bit.

grahamc:
Also, I just merged #187 -- can you rebase on top of it?

grahamc:
I think we need to revisit the labels, actually, after you made a good point about there being multiple "current" channels.

grahamc:
I think a better approach is to give each channel entry a "status" field which is `rolling`, `beta`, `stable`, `eol`. Then, set `current` from if the status is not equal to "eol".

channels.nix
```diff
@@ -1,87 +1,114 @@
...
-      "nixos-21.11-aarch64" = {
-        job = "nixos/release-21.11-aarch64/tested";
-        current = true;
-      };
+    # "Channel name" = {
+    #   job = "project/jobset/jobname"; # like https://hydra.nixos.org/job/<value>/latest-finished
+    #   # When adding a new version, determine if it needs to be tagged as a
+    #   # variant -- for example, nixos-21.05, nixos-21.05-small,
+    #   # nixos-21.05-aarch64, and nixpkgs-21.05-darwin would be variants
+    #   # "primary", "small", "aarch64", and "darwin", respectively.
```
grahamc, line 8:
```suggestion
    #   # variant -- for example:
    #   # nixos-xx.xx         => primary
    #   # nixos-xx.xx-small   => small
    #   # nixos-xx.xx-darwin  => darwin
    #   # nixos-xx.xx-aarch64 => aarch64
```

channels.nix
```diff
@@ -1,87 +1,114 @@
...
+    #   variant = "primary";
+    #   # When adding a new version, determine the status of the channel -- for
+    #   # example, unreleased NixOS versions would be "beta", released NixOS
+    #   # versions would be "stable", NixOS versions that are still supported
+    #   # but no longer the latest stable would be "deprecated", rolling
+    #   # release channels would be "rolling", channels that are no longer
+    #   # maintained but not yet removed from this list would be marked as
+    #   # "unmaintained".
```
grahamc, line 16:
```suggestion
    #
    #   # Channel Status:
    #   # '*-unstable' channels are always "rolling"
    #   # Otherwise a release progresses through phases:
    #   #
    #   #  - Directly after branch off                   => "beta"
    #   #  - Once the channel is released                => "stable"
    #   #  - Once the next channel is released           => "deprecated"
    #   #  - 3 months after the next channel is released => "unmaintained"
```

delft/eris/channel-exporter.py
```diff
@@ -70,9 +70,12 @@ def measure_channel(name):
             measurement = measure_channel(channel)
             if measurement is not None:
                 revision = measurement['revision']
+                stability = about.get('stability', '')
+                variant = about.get('variant', '')
+                current = int(about['stability'] not in ['unmaintained', 'deprecated'])
```
grahamc, line 75: I think a "deprecated" channel is still current?

channels.nix
```diff
@@ -1,87 +1,114 @@
...
+      stability = "unmaintained";
+    };
+    "nixpkgs-20.03-darwin" = {
+      job = "nixpkgs/nixpkgs-20.03-darwin/darwin-tested";
+      variant = "darwin";
+      stability = "unmaintained";
+    };
   };
 
   channels-with-urls = (builtins.mapAttrs (name: about: about.job) channels);
```
grahamc, line 114: Should we validate the variant and stability match the expected values here?

grahamc (commented):
This is looking pretty good. I left some feedback about documentation to make it easier to see the values at a glance. What do you think?

channels.nix
cole-h: I think that, since the "bounds" are described in this file, it's fine to not validate. Also, it's perfectly valid that `variant` doesn't exist (e.g. in the case of `nixpkgs-unstable`).

channels.nix
```diff
@@ -1,87 +1,122 @@
...
+    #
+    #   # Channel Status:
+    #   # '*-unstable' channels are always "rolling"
+    #   # Otherwise a release generally progresses through the following phases:
+    #   #
+    #   #  - Directly after branch off                   => "beta"
+    #   #  - Once the channel is released                => "stable"
+    #   #  - Once the next channel is released           => "deprecated"
+    #   #  - N months after the next channel is released => "unmaintained"
+    #   #    (check the release notes for when this should happen)
```
cole-h, line 24: While this *should* be one month, this might change down the road. The source of truth for that specific release should be its release notes.

PR #194, opened 2021-12-17
delft/eris: update disk monitoring alerts

mweinelt:
Quite recently the `/scratch` partition on bastion ran out of free disk
space.

This is an update to the alerting rule to generalize and extend the
existing disk fullness alerts.

In addition to monitoring `/scratch` we are now sharing the filesystem
selector between these rules and in addition monitor relative fullness
on them.

### DIFF
--- a/delft/eris.nix
+++ b/delft/eris.nix
@@ -130,10 +130,15 @@ in {
 
         {
           name = "system";
-          rules = [
+          rules = let
+            diskSelector = ''mountpoint=~"(/|/scratch)"'';
+          in
+          [
             {
-              alert = "RootPartitionLowInodes";
-              expr = ''node_filesystem_files_free{mountpoint="/"} <= 10000'';
+              alert = "PartitionLowInodes";
+              expr = ''
+                node_filesystem_files_free{${diskSelector}} <= 10000
+              '';
               for = "30m";
               labels.severity = "warning";
               annotations.summary = "{{ $labels.device }} mounted to {{ $labels.mountpoint }} ({{ $labels.fstype }}) on {{ $labels.instance }} has {{ $value }} inodes free.";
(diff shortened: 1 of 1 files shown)


delft/eris.nix
```diff
@@ -130,19 +130,28 @@ in {
...
               annotations.summary = "{{ $labels.device }} mounted to {{ $labels.mountpoint }} ({{ $labels.fstype }}) on {{ $labels.instance }} has {{ $value }} inodes free.";
               annotations.grafana = "https://monitoring.nixos.org/grafana/d/5LANB9pZk/per-instance-metrics?orgId=1&refresh=30s&var-instance={{ $labels.instance }}";
             }
 
             {
-              alert = "RootPartitionLowDiskSpace";
-              expr = ''node_filesystem_avail_bytes{mountpoint="/"} <= 10000000000'';
+              alert = "PartitionLowDiskSpace";
+              expr = ''
+                (node_filesystem_avail_bytes{${diskSelector}} * 10^(-9) <= 10)
```
mweinelt, line 151: Easier to reason about if we write it like that.

delft/eris.nix
```diff
@@ -130,19 +130,28 @@ in {
...
+                or
+                (((node_filesystem_avail_bytes{${diskSelector}} / node_filesystem_size_bytes) * 100) <= 10)";
```
mweinelt, line 153: The complete expression should always return the value of the first expression, so the actual available size, even if the percental values matches.

grahamc:
I'm getting:

```
Checking /nix/store/ycldz0n6qyalsaxikwx77fsgkdy2qf0f-rules-checkrules-checked
  FAILED:
/nix/store/ycldz0n6qyalsaxikwx77fsgkdy2qf0f-rules-checkrules-checked: 1:1242: group "system", rule 2, "PartitionLowDiskSpace": could not parse expression: 3:103: parse error: unterminated quoted string

error: builder for '/nix/store/mxyfklr0bbcpdvplm5kc0grsa4hrfknf-rules-checkrules-checked.drv' failed with exit code 1;
       last 4 log lines:
       > Checking /nix/store/ycldz0n6qyalsaxikwx77fsgkdy2qf0f-rules-checkrules-checked
       >   FAILED:
       > /nix/store/ycldz0n6qyalsaxikwx77fsgkdy2qf0f-rules-checkrules-checked: 1:1242: group "system", rule 2, "PartitionLowDiskSpace": could not parse expression: 3:103: parse error: unterminated quoted string
       >
       For full logs, run 'nix log /nix/store/mxyfklr0bbcpdvplm5kc0grsa4hrfknf-rules-checkrules-checked.drv'.

```

delft/eris.nix
grahamc: Looks like the `";` are a mistake here, I'll try deleting those

grahamc:
Sorted it!

PR #195, opened 2021-12-17
Reuse prometheus-exporters from nixpkgs

mweinelt:
These exporters have been tracked and upgraded within this repository, but can easily be reused from nixpkgs.


delft/eris.nix
```diff
@@ -568,22 +566,14 @@ in {
...
-
-    script = ''
-      export FASTLY_API_TOKEN=$(cat /run/keys/fastly-read-only-api-token)
-      ${fastly}/bin/fastly-exporter \
-        -endpoint http://127.0.0.1:9118/metrics
-    '';
+  services.prometheus.exporters.fastly = {
+    enable = true;
+    listenAddress = "127.0.0.1";
+    tokenPath = /run/keys/fastly-read-only-api-token;
```
cole-h, line 572:
```suggestion
    tokenPath = "/run/keys/fastly-read-only-api-token";
```

mweinelt: I wondered whether that would be problematic, but the module says `null` or `path`.

https://github.com/NixOS/nixpkgs/blob/master/nixos/modules/services/monitoring/prometheus/exporters/fastly.nix#L23

cole-h: I've wondered the same in the past, but don't worry!

```
vin@scadrial ~> nix repl '<nixpkgs>'
Welcome to Nix 2.5pre20211119_9cd8cff. Type :? for help.

Loading '<nixpkgs>'...
trace: Using <nixpkgs> compat wrapper...
Added 14334 variables.

nix-repl> lib.types.path.check "/asdf"
true

nix-repl> lib.types.path.check "/run/keys/fastly-read-only-api-token"  
true
```

mweinelt: Awesome. Thanks for the demonstration. Pushed a fixed version.

PR #199, opened 2022-01-29
channels: add nixos-21.11-x86_64

wamserma:
Add the x86_64 channel complementary to aarch64.

While the aggregate channel is important to keep heterogeneous system
landscapes in sync, there is still a substantial user base that only
cares about x86_64.

However, the aggregate channel is often blocked from advancing by build/test
failures that only affect aarch64. As x86_64 is a tier 1 platform (as per RFC0046),
it deserves a dedicated channel.

The impact on Hydra should be negligible, due to caching and no new build artefacts
being introduced.

There has been also been some discussion around this, e.g. in https://github.com/NixOS/nixpkgs/issues/150532
and occasionally on Discourse and so far no one opposed to creating this channel.

### DIFF
--- a/channels.nix
+++ b/channels.nix
@@ -3,7 +3,7 @@ rec {
     # "Channel name" = {
     #   # This should be the <value> part of
     #   # https://hydra.nixos.org/job/<value>/latest-finished
-    #   job = "project/jobset/jobname"; 
+    #   job = "project/jobset/jobname";
     #
     #   # When adding a new version, determine if it needs to be tagged as a
     #   # variant -- for example:
@@ -57,6 +57,11 @@ rec {
     "nixos-21.11-aarch64" = {
       job = "nixos/release-21.11-aarch64/tested";
       variant = "aarch64";
+      status = "stable";
+    };
+    "nixos-21.11-x86_64" = {
+      job = "nixos/release-21.11-x86_64/tested";
+      variant = "x86_64";
       status = "stable";
     };
 


wamserma:
ping @vcunat @grahamc

vcunat:
> The impact on Hydra should be negligible, due to caching and no new build artefacts
being introduced.

I'm actually not convinced that's true.  To me, each large NixOS jobset evaluation seems to be quite expensive in practice.  I commonly see the machines spending hours just by copying data here and there, just to build all the `/etc` trees (and similar) needed to run all the VM tests.

wamserma:
The syncing issue is already present with the aarch64 channel.

I'm a bit lost finding documentation on how the jobsets get triggered on Hydra and whether the main release-channel can be defined in terms of the architecture-specific channels. (As a side note: it also seems strange to me that the aarch64-channel carries x86 ISOs and OVAs).

vcunat:
https://channels.nixos.org/?prefix=nixos-21.11-aarch64/ contains lots of non-sensical dead links.  (e.g. we never had i686 plasma ISO)

vcunat:
Jobsets have configuration for "Check interval".  And evaluations can get triggered manually as well.

samueldr:
The isos on the "releases" page are handled by a script, and does not reflect the jobsets. A few outputs have been cherry-picked.

See this previous discussion:

 - https://github.com/NixOS/nixos-channel-scripts/pull/25#issuecomment-521511375

wamserma:
I think https://github.com/NixOS/nixos-org-configurations/blob/master/terraform/s3_listing.html.tpl is involved at least partially in the dead links.

But that should really become a new issue/PR. This one is about adding the x86_64 channel.

wamserma:
ping @grahamc

grahamc:
I'm not keen on adding this jobset today for the reasons @vcunat have mentioned. The evaluation would be heavy. I find the idea that Tier-1 archs deserve their own independent channels is an interesting idea, but I'm not sure I would read that in to the definition of the tiers. I do think a lot of good would come out of defining the goals and use cases of the channels and defining specific outcomes we want for our users.

PR #201, opened 2022-02-01
channels: 21.05 is unmaintained

vcunat:
The promise was until the end of 2021, and there's almost
no backporting to 21.05 anymore.


vcunat:
I can't see it changing on status.nixos.org.  IIRC there's usually a missing step in redeploying `bastion`?

mweinelt:
The change is still not visible.

PR #202, opened 2022-02-10
survey.nixos.org now running limesurvey

garbas:


garbas:
This was already deployed to survey.nixos.org.

Issue #207, opened 2022-02-20
builds using specific .nar file from cache.nixos.org fail with 503


busti:
**Is your feature request related to a problem? Please describe.**
I tried rebuilding my system today, first updating the channel and then running `nixos-rebuild switch`  
This resulted in the following error message:
```
unable to download 'https://cache.nixos.org/nar/0g3mvx4rg81g9fdcjc5822v14vf73lnr84fcbxa8jdgciqa1m3qk.nar.xz': HTTP error 503
cannot build derivation '/nix/store/y2qkr4rzwxkghvs2ld9p4gsapknnknj8-cudatoolkit-10.2.89.drv': 1 dependencies couldn't be built
cannot build derivation '/nix/store/987vs37cvg7ryjps6z3vchv02rzfdbjn-nvtop-1.2.2.drv': 1 dependencies couldn't be built
```
Following the link with a browser reveals the following http server error message:  
> # Error 503 Response object too large
> Response object too large
> 
> Guru Mediation:
> Details: cache-hhn4082-HHN 1645394518 540221339
> ---
> Varnish cache server

**Describe the solution you'd like**
Disappearance of the error message and successful download of the file.

**Describe alternatives you've considered**
Downgrading my system channel or switching to unsable. I tried both and it does not work, the same error appears.  
I also tried searching for mirrors of cache.nixos.org, but there seem to be no up-to-date ones.  
Many other package managers seem to have this, i.e. apt has lot's of mirrors hosted by universities, but this does not seem to be a thing with nixos.

**Additional context**
I ran the scripts provided at the landing page of `cache.nixos.org` 
Below is their result:  
```
domain=cache.nixos.org
> dig -t A cache.nixos.org
    
    ; <<>> DiG 9.16.25 <<>> -t A cache.nixos.org
    ;; global options: +cmd
    ;; Got answer:
    ;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 39869
    ;; flags: qr rd ra; QUERY: 1, ANSWER: 2, AUTHORITY: 0, ADDITIONAL: 1
    
    ;; OPT PSEUDOSECTION:
    ; EDNS: version: 0, flags:; udp: 1232
    ;; QUESTION SECTION:
    ;cache.nixos.org.		IN	A
    
    ;; ANSWER SECTION:
    cache.nixos.org.	1509	IN	CNAME	dualstack.v2.shared.global.fastly.net.
[...]

PR #208, opened 2022-03-17
terraform: enable segment caching for NAR files

zimbatm:
This should fix the issue we're seeing in #207

### DIFF
--- a/terraform/cache.tf
+++ b/terraform/cache.tf
@@ -197,6 +197,20 @@ resource "fastly_service_v1" "cache" {
     type     = "recv"
   }
 
+  # Work around the 2GB size limit for large files
+  #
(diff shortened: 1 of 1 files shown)


vcunat:
The docs look good to me: https://docs.fastly.com/en/guides/segmented-caching

zimbatm:
> HTTP chunked transfer encoding between Fastly and origin isn't supported. Your origin server must frame responses to Range: requests with the Content-Length heade

that's my only concern, but I think S3 behaves properly

vcunat:
Right, in this part I only hoped; I don't know such HTTP details.

vcunat:
Yeah, maybe `.nar.xz` is parsed as just `.xs` extension?

zimbatm:
okay, using the path matching works, and is more future-proof

PR #213, opened 2022-07-19
Add live.nixos.org DNS entry

bryanhonof:
For the Summer of Nix Lecture Series we set up an Owncast instance,
currently running under live.bjth.xyz. I'd be better marketing if we ran
that under live.nixos.org instead :).

### DIFF
--- a/terraform/dns.tf
+++ b/terraform/dns.tf
@@ -188,8 +188,18 @@ locals {
     },
     {
       hostname = "live.nixos.org"
+      type     = "TXT"
+      value    = "machine-owner=@bryanhonof provider=hetzner-cloud service=owncloud"
+    },
+    {
+      hostname = "live.nixos.org"
       type     = "A"
       value    = "167.235.51.204"
+    },
+    {
+      hostname = "live.nixos.org"
+      type     = "AAAA"
+      value    = "2a01:4f8:c0c:96cc::42"
     },
   ]
 }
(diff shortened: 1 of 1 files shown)


terraform/dns.tf
```diff
@@ -186,6 +186,11 @@ locals {
       type     = "TXT"
       value    = "v=spf1 include:spf.improvmx.com ~all"
     },
+    {
+      hostname = "live.nixos.org"
+      type     = "A"
```
SuperSandro2000, line 191: Where is IPv6?

bryanhonof: It currently has none. Will add one if this is necessary.

zimbatm (commented):
One of the issues we are facing as the infra team is how to keep track and contact the various owners of infrastructure. Some argument could also be made around security and attack surfaces.

I think as long as it's temporary, it's fine, but if the instance stays around, we should consider transferring the machine and configuration to the infra team.

I like the initiative and enthusiasm so please don't take this as a criticism :)

bryanhonof:
@zimbatm 
> I think as long as it's temporary, it's fine, but if the instance stays around, we should consider transferring the machine and configuration to the infra team.

We were thinking about also using either that machine, or just the setup, for NixCon 2022. We're testing out the setup during this year's  Summer of Nix. After those events the platform could still be used to stream though, but I don't think it'd need to be online 24/7.

I do agree that it'd be better if the machine, and configuration, were in hands of the infra team. We're currently still playing, and discovering, it ourselves 😄. What do you think the best way forward is?

zimbatm (approved):

zimbatm:
For now, I applied the changes, and we can revisit that question after NixCon is over.

PR #226, opened 2022-11-17
add NixOS/nixos-metrics metrics via VictoriaMetrics

Radvendii:
The metrics we've been collecting at github:NixOS/nixos-metrics are currently viewable at https://nixos.github.io/nixos-metrics/. It would be nice to get them into the community Grafana, so people will know where to find them and so they can be accessed using all of Grafana's fancy features (data analysis, comparison to other metrics, etc).

We are aiming to use VictoriaMetrics rather than Prometheus because
1. It allows us to add more metrics downstream that will automatically get picked up without making a PR here every time
2. It allows back-filling the data we've already collected.

I'm not sure if there's an easy way to allow external yet secure access directly to the VictoriaMetrics instance. So instead I've opted for the server to *pull* data and push it to VictoriaMetrics internally.

**NOTE:** There is one manual step required. VictoriaMetrics has to be added as a source in Grafana of type Prometheus and with the address `http://localhost:8428/prometheus`

### DIFF
--- a/delft/eris.nix
+++ b/delft/eris.nix
@@ -590,6 +590,28 @@ in
       (import ../channels.nix).channels;
   };
 
+  # pull nixos metrics from github:NixOS/nixos-metrics to local VictoriaMetrics
+  # every day at 09.00
+  services.cron = {
+    enable = true;
+    systemCronJobs = let
+      inherit (config.services.victoriametrics) listenAddress;
+      importURL = "http://localhost:${listenAddress}/api/v1/import";
+      resetURL = "http://localhost:${listenAddress}/internal/resetRollupResultCache";
+      dataURL = "https://raw.githubusercontent.com/NixOS/nixos-metrics/data/victoriametrics.jsonl";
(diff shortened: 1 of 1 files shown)


delft/eris.nix
```diff
@@ -590,6 +590,28 @@ in
       (import ../channels.nix).channels;
   };
 
+  # pull nixos metrics from github:NixOS/nixos-metrics to local VictoriaMetrics
+  # every day at 09.00
+  services.cron = {
```
edolstra, line 595: This should be a systemd timer. Cron is a pain in the ass to use/test.

Radvendii: Fixed.

thufschmitt:
@grahamc : Eelco told me you've deployed this but I can't find these metrics anywhere on https://monitoring.nixos.org/grafana . Any pointer as to where this is?

mweinelt:
@thufschmitt The data source was never added to Grafana.

PR #232, opened 2022-12-16
enable TLS redirect for hydra.nixos.org

yu-re-ka:
In 2022 one should not accidentally be on a non-TLS website when just entering "hydra.nixos.org" in the url bar


grahamc:
Yikes, thanks!

PR #233, opened 2023-01-01
channels: Mark 22.05 unmaintained

mweinelt:
EOL was on 2022-12-31.

### DIFF
--- a/channels.nix
+++ b/channels.nix
@@ -75,69 +75,6 @@ rec {
       variant = "aarch64";
       status = "unmaintained";
     };
-
-    "nixos-21.11" = {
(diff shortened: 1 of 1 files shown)


vcunat:
Maybe also drop 20.09 completely?

mweinelt:
Fine with me. What about 21.05 and 21.11?

vcunat:
Not sure.  Last updated 5 months ago so perhaps effectively dead anyway.

vcunat:
Missing deploy to `bastion`?  I still can't see any difference on https://status.nixos.org/

PR #234, opened 2023-01-24
Introduce reproducible.nixos.org

raboof:
It would be good to have a place to collect information about Reproducible Builds on Nix. Currently there is https://r13y.com, but it is rather outdated and it would be good to gradually move this to more shared infrastructure so we can collaborate on maintaining it.

This is a first step in that direction, having more up-to-date reports hosted from my uberspace.de account. Next steps might be to generate those reports from nix-community infra and host them on the Nix netlify, but let's take one step at a time.

You can preview the content at https://nix-rb.bzzt.net

### DIFF
--- a/terraform/dns.tf
+++ b/terraform/dns.tf
@@ -72,6 +72,11 @@ locals {
       value    = "2a01:4f8:c0c:6e2c::1"
     },
     {
+      hostname = "reproducible.nixos.org"
+      type     = "CNAME"
+      value    = "nixos.github.io"
+    },
+    {
       hostname = "_293364b7f7ebb076ac287cd132f8b316.cache.ngi0.nixos.org"
       type     = "CNAME"
       value    = "_6a75cfb0c20f4eaac96b72afaffb489b.auiqqraehs.acm-validations.aws"


0x4A6F (approved):
Works and it is a good idea to put this under the umbrella of nixos.org!
```
curl -k --resolve reproducible.nixos.org:443:185.26.156.130 https://reproducible.nixos.org/
curl -k --resolve reproducible.nixos.org:443:185.26.156.130 https://reproducible.nixos.org/nixos-iso-minimal-r13y/
curl -k --resolve reproducible.nixos.org:443:185.26.156.130 https://reproducible.nixos.org/nixos-iso-gnome-r13y/
```

bryanhonof:
This looks great to me!

@grahamc since you have r13y.com, what do you think?

zimbatm:
@raboof you are now part of the @NixOS/reproducible team, and you can invite people you think are relevant to your work. As part of this team, you can also maintain the NixOS/reproducible.nixos.org repo to put the work you did on r13y.

One thing we're trying to avoid is having the infrastructure split all over the place, and then forgetting who has access to what.

zimbatm:
Sounds good, let's see how far we can take this :)

Atemu:
@grahamc could you make r13y.com redirect to reproducible.nixos.org?

Artturin:
the links on https://github.com/NixOS/nixpkgs/pull/224670 don't work

Artturin:
the links on http://reproducible.nixos.org/nixos-iso-gnome-r13y/ don't work (404)

raboof:
> the links on http://reproducible.nixos.org/nixos-iso-gnome-r13y/ don't work (404)

Let's discuss this in the https://github.com/NixOS/reproducible.nixos.org repo - I created https://github.com/NixOS/reproducible.nixos.org/issues/3

PR #235, opened 2023-01-25
eris: fix go-neb secret ownership

lheckemann:
This will probably fix matrix alerts.


zimbatm:
what is go-neb? :)

lheckemann:
It's a matrix bot; we only use it for alerts (https://matrix.to/#/#infra-alerts:nixos.org) currently, but these are broken because of the secret ownership.

PR #240, opened 2023-03-01
Add live.nixos.org records

bryanhonof:
For Nix's 20th anniversary, the marketing team had the idea to host livestreams like during the Summer of Nix. I've just re-used the server config of the Summer of Nix, but these records were removed back then.

cc: @zimbatm @garbas


zimbatm (approved):
sounds good!

PR #242, opened 2023-03-08
mirror-tarballs: partial fixing of the service

vcunat:


vcunat:
(deployed already but apparently still needs further fixes to succeed)

Artturin:
Maybe merge this if the changes here work even if the uploading doesn't work

vcunat:
With this the git parts look OK to me, but the AWS part still breaks with permission access, but that should be mostly independent.

PR #244, opened 2023-04-03
rhea: add hydra-scale-equinix-metal

lheckemann:


grahamc:
Thanks for getting this over the line, it sucked that I wrote the autoscaler work and then never finished the job.

PR #246, opened 2023-04-05
rfc39: update github.com ssh key

vcunat:


vcunat:
The background, in case you haven't noticed:
https://github.blog/2023-03-23-we-updated-our-rsa-ssh-host-key/

modules/rfc39.nix
```diff
@@ -28,7 +28,8 @@ in {
   users.groups.rfc39 = {};
 
 
-  programs.ssh.knownHosts."github.com".publicKey = "ssh-rsa AAAAB3NzaC1yc2EAAAABIwAAAQEAq2A7hRGmdnm9tUDbO9IDSwBK6TbQa+PXYPCPy6rbTrTtw7PHkccKrpp0yVhp5HdEIcKr6pLlVDBfOLX9QUsyCOV0wzfjIJNlGEYsdlLJizHhbn2mUjvSAHQqZETYP81eFzLQNnPHt4EVVUh7VfDESU84KezmD5QlWpXLmvU31/yMf+Se8xhHTvKSCZIFImWwoG6mbUoWf9nzpIoaSjB+weqqUUmpaaasXVal72J+UX2B+2RPW3RcT0eOzQgqlJL3RKrTJvdsjE3JEAvGq3lGHSZXy28G3skua2SmVi/w4yCE6gbODqnTWlg7+wC604ydGXA8VJiS5ap43JXiUFFAaQ==";
+  programs.ssh.knownHosts."github.com".publicKey =
```
SuperSandro2000, line 31: I am not fully sure how ssh chooses which host key is used but I would just add both ed25519 and rsa to be safe.

mweinelt: Any key will do, sticking with just the ed25519 one is fine.

vcunat: Well, it works.

PR #254, opened 2023-06-02
rewrite nixpkgs's create-amis script using Terraform

AmineChikhaoui:
This rewrites [create-amis.sh]( https://github.com/NixOS/nixpkgs/blob/master/nixos/maintainers/scripts/ec2/create-amis.sh) using terraform.
The create-amis.sh was pretty slow and switching to Terraform makes things more declarative and easier to read.
There are few things I don't like in the Terraform code, since we're copying the AMI to a set of regions (which require a different aws provider per region), the module system of terraform doesn't really work for this. The easy solution was to just use `jq` and `make` to generate the copy ami resources in JSON spec instead of HCL.
Even using Terraform, this still takes 15minutes which isn't that fast either but at least ami copying is done in parallel.

A note about usage of Terraform workspaces, since we don't want to trigger terraform replacements i.e delete and recreate of resources as we want to keep AMIs being used (and perhaps use a lifecycle to purge them periodically), I went with a workspace (statefile) per `{release}.{arch}.{buildId}`.

Things I'm planning to explore now that the building blocks are here:

- [ ] Add an example terraform+flake ec2 instance config using the generated AMIs as a quick way to test.
- [x]  Use Github Actions for running the Terraform applies. Explore OIDC integration with AWS: https://docs.github.com/en/actions/deployment/security-hardening-your-deployments/configuring-openid-connect-in-amazon-web-services. 
- [ ] Setup some lifecycle policies to purge older AMIs, which will allow scheduling this in cron without worrying about accumulated costs from snapshots, objects, etc ..


AmineChikhaoui:
@zimbatm @grahamc if you can do a quick review or if there are any suggestions.

AWS OIDC setup is pointing to my fork/branch but can switch it before merging to nixos/nixos-org-configurations. I can also terraform the IAM role config for visibility.

Here is an example of the workflow runs: https://github.com/AmineChikhaoui/nixos-org-configurations/actions/runs/5172652543

cole-h:
Something that I just discovered is the `aws_dlm_lifecycle_policy` resource, which has an example for cross-region copy: https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/dlm_lifecycle_policy#example-cross-region-snapshot-copy-usage. Supposedly, setting [`policy_type`](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/dlm_lifecycle_policy#policy_type) to `IMAGE_MANAGEMENT` and [`resource_types`](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/dlm_lifecycle_policy#resource_types) to `INSTANCE` might work? Might be worth looking into, if it means we might be able to get rid of the jq and scripted processing.

Though I don't know if the cross-region copying means you'd need a provider for each region (as is currently necessary). I also don't know if the fact that it's for managing a lifecycle makes it unsuitable for this, though (i.e. if it _requires_ you to specify something that would "clean up" images / etc and not just use it for cross-region copying).

AmineChikhaoui:
> EDIT2: There is also this [`aws_ami_copy`](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/ami_copy) resource that might be useful. I think it still requires a separate provider for each region we want to copy _to_, but...

That's the one I'm using in [regions.jq](https://github.com/NixOS/nixos-org-configurations/pull/254/files#diff-74b39e5d0cf00e3b27d66c02c4f198f5d2a3698934d664175f9419b6425ad0a1R10) right ?

AmineChikhaoui:
@cole-h probably CDK but not sure if it's worth the effort compared to a few lines of `jq` file 😄 

I'll look into DLM, I was planning to explore it for managing the deprecation and removal of AMIs. Didn't see copy AMI handling specifically from a quick scroll over docs but can look a bit more.

delroth:
Trying to understand the status of this PR - what is the relation between this work and the (current operational, I think) https://github.com/NixOS/amis maintained by @arianvp? Are these complementary / parallel implementations / does one deprecate the other / ...?

(Bear with me, I have very little context, but I'd like to clean up this repo and make sure we have a clear set of pending items to look at :p)

AmineChikhaoui:
@delroth I didn't read through https://github.com/NixOS/amis yet but my understanding it will deprecate this process and this PR won't be needed anymore in that case.

mweinelt:
The [nixos/amis](https://github.com/NixOS/amis) project is in production for a while now, so this PR has indeed been obsoleted.

PR #263, opened 2023-08-05
non-critical-infra: hetzner-01 bootstraping

JulienMalka:
Closes https://github.com/NixOS/nixos-org-configurations/issues/260
I have bootstrapped the machine for the non-critical infra team. 
Added keys for @zimbatm, @mweinelt and myself so far. This PR is for information (because the machine has de facto the configuration described here) and I will merge it after 24h.

### DIFF
--- a/non-critical-infra/hosts/hetzner-01/default.nix
+++ b/non-critical-infra/hosts/hetzner-01/default.nix
@@ -1,0 +1,64 @@
+{ config, pkgs, lib, inputs, ... }:
+
+{
+  imports =
+    [
+      ./hardware.nix
+      inputs.srvos.nixosModules.server
+      inputs.srvos.nixosModules.hardware-hetzner-online-amd
+    ];
+
+
+  deployment = {
+    targetHost = "65.109.26.213";
+    targetUser = "root";
+  };
+
+  # Bootloader.
+  boot.loader.grub.enable = true;
+  boot.supportedFilesystems = [ "zfs" ];
+  boot.loader.grub.mirroredBoots = [
+    { path = "/boot-1"; devices = [ "/dev/disk/by-id/nvme-SAMSUNG_MZQL23T8HCLS-00A07_S64HNJ0T508051" ]; }
+    { path = "/boot-2"; devices = [ "/dev/disk/by-id/nvme-SAMSUNG_MZQL23T8HCLS-00A07_S64HNJ0T508053" ]; }
+  ];
+  boot.loader.grub.useOSProber = true;
+
+
+  networking.hostName = "hetzner-01"; # Define your hostname.
+  networking.hostId = "745b334a";
+
+  disko.devices = import ./disko.nix;
+
+  # Set your time zone.
+  time.timeZone = "UTC";
+
+  environment.systemPackages = with pkgs; [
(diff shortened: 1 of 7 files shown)


non-critical-infra/hosts/hetzner-01/default.nix
```diff
@@ -0,0 +1,64 @@
...
+
+  services.openssh.enable = true;
+
+  networking.firewall.allowedTCPPorts = [ ];
+  networking.firewall.allowedUDPPorts = [ ];
+
+  systemd.network.networks."10-uplink".networkConfig.Address = "2a01:4f9:5a:186c::2";
+
+  users.users.root.openssh.authorizedKeys.keyFiles = [
+    (pkgs.fetchurl {
```
cole-h, line 47: IMHO these should just be strings of the keys, rather than fetching from a floating URL.

non-critical-infra/hosts/hetzner-01/default.nix
```diff
@@ -0,0 +1,64 @@
...
+    ];
+
+
+  deployment = {
+    targetHost = "65.109.26.213";
+    targetUser = "root";
+  };
+
+  # Bootloader.
+  boot.loader.grub.enable = true;
```
cole-h, line 18: A bit of a shame to be using grub. Does the box not support UEFI boot (systemd-boot)? (Not that it's worth changing now, I suppose.)

non-critical-infra/hosts/hetzner-01/default.nix
```diff
@@ -0,0 +1,64 @@
...
+  boot.loader.grub.useOSProber = true;
+
+
+  networking.hostName = "hetzner-01"; # Define your hostname.
+  networking.hostId = "745b334a";
+
+  disko.devices = import ./disko.nix;
+
+  # Set your time zone.
+  time.timeZone = "Europe/Paris";
```
0x4A6F, line 33: Any reason for setting this to your local time and not to UTC? You could set TZ in your environment to `Europe/Paris` instead.

0x4A6F (commented):
Thanks for your work on this.

non-critical-infra/hosts/hetzner-01/default.nix
Mic92: I'd also set time for all machines to the same timezone. It makes later things like correlating logs easier.

non-critical-infra/hosts/hetzner-01/default.nix
JulienMalka: I think these boxes are using bios by default, and uefi could be activated if we ask.

non-critical-infra/hosts/hetzner-01/default.nix
JulienMalka: Yeah sorry that's indeed a miss on my side, I'll change that and redeploy.

non-critical-infra/hosts/hetzner-01/default.nix
Mic92: Newer machines now have uefi by default on hetzner.

zimbatm:
thanks!

PR #271, opened 2023-09-05
Move releases to Infrequent Access after a year

edolstra:
Since we have 28 TB in this bucket, this gives a non-trivial savings.


zimbatm (approved):
good idea

PR #274, opened 2023-09-10
Reuse ssh-keys for non-critical-infra

mweinelt:
Creates the `infra` key list and migrates hetzner-01 to run with it.

The wrong keys were deployed for me, so I had to access the host using the keypair I use for GitHub.

Fetching keys from GitHub is not great for key hygiene and did in fact cause a hash mismatch on Julien's keys.

@JulienMalka Please set up keys in `./ssh-keys.nix`, deploying all 14 keys from GitHub feels wrong.


JulienMalka:
You're right, it's better this way

PR #275, opened 2023-09-10
hetzner-01: rename to caliban.nixos.org

mweinelt:
We needed a name to create CNAMES to eventually, so here is one proposal, since we're going for pet names.

Caliban is the second largest moon of Uranus, and should somewhat fit into the existing naming scheme. I'm open to alternatives.

https://en.wikipedia.org/wiki/Caliban_(moon)

Once we have the DNS name deployed we can update the deployment address accordingly.

cc https://github.com/NixOS/nixos-org-configurations/issues/269#issuecomment-1693723824


JulienMalka:
That's fine by me.

zimbatm (approved):
thanks for proposing a naming scheme so we don't have to bikeshed on this :)

mweinelt:
I took rhea, haumea and eris and ran with it!

zimbatm:
applied!

PR #279, opened 2023-09-19
add dns record for vault.nixos.org

JulienMalka:
In preparation for an upcoming PR deploying vaultwarden

### DIFF
--- a/terraform/dns.tf
+++ b/terraform/dns.tf
@@ -250,6 +250,11 @@ locals {
       hostname = "caliban.nixos.org"
       type     = "A"
       value    = "65.109.26.213"
+    },
+    {
+      hostname = "vault.nixos.org"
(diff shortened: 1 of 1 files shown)


mweinelt:
Why did you choose the plural form?

JulienMalka:
Hmm I don't know, a vaultwarden instance may host several vaults that can be shared with different people in function of the need, but I don't have any strong feeling.

zimbatm:
I can see myself using the singular instead of the plural by mistake. But you get to pick since you're deploying it. It could also be pass.nixos.org, bitwarden.nixos.org or bw.nixos.org.

Let me know :)

JulienMalka:
Changed to `vault.nixos.org`, less confusing :)

JulienMalka:
Thanks

PR #283, opened 2023-10-03
terraform: access AWS using SSO

zimbatm:
Replace the long-lived AWS key-pair with AWS SSO / Identity Center.

This will allow us to:
* remove the IAM users in the old account
* tightening up the security by using short-lived credentials
* provide access to some of the resources to the new team


terraform/terraform.tf
```diff
@@ -4,6 +4,7 @@ terraform {
     encrypt = true
     key     = "targets/terraform"
     region  = "eu-west-1"
+    profile = "nixos-prod"
```
cole-h, line 7: I'm not that knowledgeable about tf profiles, but it seems weird to me that this is in the `nixos-prod` profile, but in `eu-west-1`, while above the `nixos-prod` profile is assigned to `us-east-1`... Maybe it doesn't matter though.

PR #286, opened 2023-10-16
public fastly logs

zimbatm:
Give access to the logs, so everybody can run their own analysis. The requester has to pay so it doesn't drain the foundation funds.

Eg:
```
aws s3 ls s3://fastly-logs-20220622145016462800000001/ --request-payer
```

### DIFF
--- a/terraform/fastlylog/main.tf
+++ b/terraform/fastlylog/main.tf
@@ -21,6 +21,12 @@ resource "aws_s3_bucket" "logs" {
   }
 }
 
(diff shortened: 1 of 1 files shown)


zimbatm:
TODO: Address Eelco's privacy concerns. I think we will have to remove User-Agent and ASN from the logs to make it public.

zimbatm:
There is much more cleaning that needs to happen before we can make that data public.

PR #290, opened 2023-10-17
cache: add parquet inventory

zimbatm:
This makes it cheaper to traverse and analyze the S3 list of files.

This will be useful to answer a few questions like:
* what is the oldest file that we have?
* how much data is stored in what time ranges?
* give me the list of all .drv or .narinfo files in the bucket.
* how many files of which type do we have?
* what is the distribution of file sizes?


zimbatm:
Already applied so we can start getting data soon

PR #293, opened 2023-10-25
iam: add new user for fastly<->cache authn

delroth:
Guessing at how things are architectured, lmk if that matches how you want the two Terraform codebases to be split and how you want data to be shared between the two.

Making this a separate PR so that @zimbatm (or someone else) can apply the diff early, and then I have an IAM identity ready to play with for the main change.

Ref #277

```
  # aws_iam_access_key.fastly-cache-access will be created
  + resource "aws_iam_access_key" "fastly-cache-access" {
      + create_date                    = (known after apply)
      + encrypted_secret               = (known after apply)
      + encrypted_ses_smtp_password_v4 = (known after apply)
      + id                             = (known after apply)
      + key_fingerprint                = (known after apply)
      + secret                         = (sensitive value)
      + ses_smtp_password_v4           = (sensitive value)
      + status                         = "Active"
      + user                           = "fastly-cache-access"
    }

  # aws_iam_user.fastly-cache-access will be created
  + resource "aws_iam_user" "fastly-cache-access" {
      + arn           = (known after apply)
      + force_destroy = false
      + id            = (known after apply)
      + name          = "fastly-cache-access"
      + path          = "/"
      + tags_all      = (known after apply)
      + unique_id     = (known after apply)
    }

Plan: 2 to add, 0 to change, 0 to destroy.

Changes to Outputs:
  + cache = (sensitive value)
```

Issue #277, opened 2023-09-13
Investigate setting "Requester Pays" on the cache S3 bucket


delroth:
The NixOS cache S3 bucket is currently publicly accessible. However, the foundation gets charged for any access to that publicly accessible bucket (bandwidth when not accessed from AWS us-east, and ops always). This has the following problems (non exhaustive list):

- A malicious actor could bleed money from the foundation.
- A non-malicious actor could write code accessing the S3 bucket (e.g. to perform a batch analysis of some sort) and accidentally end up charging all the analysis cost to the foundation. Them being shielded from the costs also means they have no incentive to make their analysis more efficient.

AWS supports a feature called "Requester Pays" for public S3 buckets, which charges the cost of the request (bandwidth + ops) to the requester. The requester needs to authenticate their request to the bucket so charges can be accounted properly, but otherwise the contents are still publicly readable.

It's currently unclear whether this can be made to work with Fastly - for example, whether a bucket can be both "Requester Pays" for unknown users, and "Owner Pays" (current config) for certain identities. This would need some digging into AWS docs, or asking someone who knows more about S3 settings :)

delroth:
I suspect Fastly could be made to work using something similar to https://docs.fastly.com/en/guides/amazon-s3#using-an-amazon-s3-private-bucket to authenticate requests with foundation-owned credentials.

zimbatm:
This has become a priority because I suspect somebody is pulling directly from the bucket, and the invoice keeps increasing.

![image](https://github.com/NixOS/nixos-org-configurations/assets/3248/26b0d2c4-6ac3-4a42-9b19-30baf2fa73ae)

On the Fastly side, the requests are relatively constant over time:

![image](https://github.com/NixOS/nixos-org-configurations/assets/3248/debc560d-94c0-4f21-b98a-7347bb518792)

delroth:
Suggested pre-rollout plan:
1. Create a test bucket set up with similar perms to the cache bucket (RO to everyone). Make it Requester Pays.
2. Create an IAM user to use for our Fastly config, attached to the Foundation's billing account.
3. Create a test Fastly service to front that test bucket. Use the Fastly docs for private buckets (linked in my initial comment) to set up authn. Figure out how to add the magic header that grants consent for billing (x-amz-request-payer). Ensure test bucket access works.
[...]

PR #294, opened 2023-10-25
dns, fastly: add cache-staging.nixos.org domain + TLS subscription

delroth:
Splitting this out into a separate PR like #293 since I don't have permissions to add new TLS subscriptions.

The aim is to copy the current Fastly cache config to cache-staging.nixos.org so we can make sure the authn VCL works before pushing it to the main cache.nixos.org config.

DNS was already pushed:
```
$ host cache-staging.nixos.org
cache-staging.nixos.org is an alias for dualstack.v2.shared.global.fastly.net.
```

cc @zimbatm 
Ref #277


zimbatm:
applied!

PR #295, opened 2023-10-27
cache-staging: add fastly service configuration and test VCL for authentication

delroth:
This authenticates Fastly requests for cache-staging.nixos.org going to the cache S3 bucket, and sets the request-payer: "requester" header (which might be required if the IAM user is not in the same project as the S3 bucket).

Deployed already and https://cache-staging.nixos.org/nix-cache-info seems to work as expected.

Note that this duplicates the Terraform config for the cache Fastly service, because Terraform is pain and there's no way to e.g. inherit from a resource to apply selective overrides. The only differences between the two services should be:
1. cache.nixos.org -> cache-staging.nixos.org
2. no S3 logging
3. the authn VCL

Ref #277

### DIFF
--- a/terraform/cache.tf
+++ b/terraform/cache.tf
@@ -254,6 +254,159 @@ resource "fastly_tls_subscription" "cache" {
   certificate_authority = "globalsign"
 }
 
+# Temporarily duplicated while testing VCL fragment for Fastly<->S3 authn. TF
+# doesn't make it particularly easy to avoid this duplication.
+#
+# TODO: remove once the Fastly<->S3 authn is deployed to the main cache
+# Fastly service.
+resource "fastly_service_vcl" "cache-staging" {
+  name        = "cache-staging.nixos.org"
+  default_ttl = 86400
+
+  backend {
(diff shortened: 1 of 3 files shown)


zimbatm (approved):
looks good. The cache.tf is getting quite large so I would split the content into a cache_staging.tf file, but that's just a matter of preference.

delroth:
Yeah, I was considering it but I'm going to remove it whenever this change is applied to the main cache config anyway (which should be a matter of days).

zimbatm:
Ok. It seems useful to keep that around if we want to try other forms of VCL tweaking.

delroth:
Agreed, but it also seems fairly easy to just re-duplicate it once someone else wants to do changes later. I'd like to avoid keeping the duplicated config for long periods of time in the repo because then we run the risk of the two configs getting out of sync, which might be "dangerous" long term (i.e. we don't test what we wanted to test, because staging/prod have some unnoticed differences).

IMO it should be either de-duplicated or temporary.

zimbatm:
Ok

PR #296, opened 2023-10-27
terraform-iam: profile a workstation for the archeologists

zimbatm:


delroth:
How do you envision the terraform vs. terraform-iam separation? I thought terraform-iam should just be for the extra-privileged stuff, but "the configuration of the archeologists workstation" doesn't seem like such a thing.

Is this temporary and meant to be split off at some point? If so, could it be documented as such? If not, I don't understand the boundaries of the split :)

zimbatm:
You and me both :)

I had limited time to unblock the archeologist team so I hacked this together. The terraform-iam/fastlylogs also has a S3 bucket in there.

I feel like this is not the right line to split a module. It would be nice if you could read the IAM roles so at least you could run a plan without it failing in the noop case. Then we could merge those back again. And maybe split the root modules on different lines, like based on the project.

zimbatm:
I'll get that cleaned later

PR #297, opened 2023-10-28
cache: authenticate Fastly<->S3 requests in preparation for Requester Pays

delroth:
Ref #277

Configuration already tested to work on cache-staging.

Not applied yet, will push to prod in ~2h if no objection.


delroth:
Pushed to prod 5min ago, Fastly live metrics suggest that nothing has blown up. Will continue monitoring throughout the day, but this seems to be working fine.

PR #298, opened 2023-10-28
ci: add a "terraform fmt" check on pushes/PRs

delroth:
Attempt at preventing spurious diffs when people like me don't know or forget that `terraform fmt` is a thing :-)

Not technically perfect since it's not using the `terraform` version from nixpkgs or the `flake.lock`, but likely good enough. If it causes problems, we can fix it later.

### DIFF
--- a/.github/workflows/terraform-fmt.yml
+++ b/.github/workflows/terraform-fmt.yml
@@ -1,0 +1,24 @@
+name: Check Terraform formatting
+
+on:
+  push:
+  pull_request:
+
+permissions:
+  contents: read
+
+jobs:
+  terraform-fmt:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+
+    - name: Check terraform/ formatting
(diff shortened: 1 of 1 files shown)


delroth:
Sample run: https://github.com/delroth/nixos-org-configurations/actions/runs/6677461339/job/18147498203

cole-h:
No judgement, but is there a reason you chose an action over running `terraform fmt -check -recursive ./terraform{,-iam}` from inside the devshell?

delroth:
One is 3 lines of YAML, the other is more than 3 lines of YAML and more than I cared about looking up how to get working :)

As mentioned, I think running "terraform fmt" from the devshell is better, feel free to supersede this PR if you want to spend the 15min figuring out how to get that done.

(Note: it probably needs to be two separate devshells for terraform and terraform-iam, since the current setup has two separate flakes for those two. But that should probably be fixed at some point too...)

zimbatm:
I'm sure we'll revisit that later, but it's going in the right direction

PR #301, opened 2023-11-03
hydra: use cache.nixos.org URL for logs instead of raw S3 URLs

delroth:
We don't use Cloudfront anymore, and Fastly should be setting the proper CORS headers.


delroth:
(Source re: CORS: https://github.com/NixOS/nixos-org-configurations/blob/master/terraform/cache.tf#L187)

PR #311, opened 2023-11-21
channels: Add NixOS 23.11 as beta

figsoda:


vcunat:
First we need the jobsets and let them finish.  Otherwise these will cause useless failure alerts IIRC.  I'm on that.

PR #316, opened 2023-12-06
makemake: Stage Hydra PR 1316

Ericson2314:
That is https://github.com/NixOS/hydra/pull/1316, preparation for broader CA derivations support in a way that does not require scheme changes.

---

This is already deployed, I am just ending this PR so reality and this PR stay in sync. Is there a better way?

CC @tomberek

### DIFF
--- a/ngi0/makemake/flake.lock
+++ b/ngi0/makemake/flake.lock
@@ -22,15 +22,16 @@
         "nixpkgs": "nixpkgs"
       },
       "locked": {
-        "lastModified": 1701446871,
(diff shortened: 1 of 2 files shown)


Ericson2314:
@mweinelt is the way the other machines are managed more automatic?

mweinelt:
No, sadly not.

Ericson2314:
I see. My goals is to do what needs to be do to get CA derivations out without rocking the boat --- I were enough hats already without getting sucked into indefinite ops duty :sweat_smile: --- but certainly we can ping @edolstra on this. I have no objection either way.

PR #328, opened 2024-01-12
wireguard-hosts: add hexa-helix

mweinelt:
I decided against adding a port, since my end is floating anyway.


mweinelt:
Tested and works.

PR #335, opened 2024-01-25
rhea: migrate from apache2 to nginx

delroth:
Currently deployed on hydra.nixos.org for your testing's convenience.

A few things weren't ported because I can't directly tell why they'd be wanted: for example, disabling compression on API endpoints (which are usually large blobs of very compressable JSON).

We might want to tweak the log settings away from the default, but I also didn't directly see a reason we'd *need* this immediately.


delft/hydra-proxy.nix
```diff
@@ -1,71 +1,35 @@
...
-      ProxyPass         /       http://127.0.0.1:3000/ retry=5 disablereuse=on
-      ProxyPassReverse  /       http://127.0.0.1:3000/
-
-      <Location />
-        SetOutputFilter DEFLATE
-        BrowserMatch ^Mozilla/4\.0[678] no-gzip\
-        BrowserMatch \bMSI[E] !no-gzip !gzip-only-text/html
-        SetEnvIfNoCase Request_URI \.(?:gif|jpe?g|png|narinfo)$ no-gzip dont-vary
-        SetEnvIfNoCase Request_URI /api/ no-gzip dont-vary
-        SetEnvIfNoCase Request_URI /download/ no-gzip dont-vary
```
SuperSandro2000, line 28: I think we want to keep the no compression for download and nar since they are already compressed and nginx will just use CPU cycles. We should also probably disable proxy buffering for those

delft/hydra-proxy.nix
```diff
@@ -1,71 +1,35 @@
...
-      { forceSSL = true;
-        enableACME = true;
-        extraConfig = ''
-          # Required by Catalyst.
-          RequestHeader set X-Forwarded-Proto https
-          RequestHeader set X-Forwarded-Port 443
-          Header always set Strict-Transport-Security "max-age=15552000"
-
-          RewriteEngine On
-          RewriteCond %{QUERY_STRING} ^query=pkgs.chromium$
```
SuperSandro2000, line 58: Is there still a chromium wiki page that looks to a rather expensive hydra search which causes lots of traffic?

delft/hydra-proxy.nix
delroth: The largest part of /download/ requests aren't already compressed. They're things like https://nixos.org/hydra/manual/ which is served straight out of https://hydra.nixos.org/build/196107287/download/1/hydra/ (for example)

The rest is getting MIME-sniffed properly by nginx to not get re-compressed.

```
$ curl -vvv -sH 'Accept-encoding: gzip' https://hydra.nixos.org/build/196107287/download/1/hydra/css/chrome.css -o/dev/null
< content-encoding: gzip

$ curl -vvv -sH 'Accept-encoding: gzip' https://hydra.nixos.org/build/246764727/download/1/nixos-system-x86_64-linux.tar.xz -o/dev/null
< content-type: application/x-xz
(no content-encoding)
```

Disabling proxy buffering is a micro-optimization which doesn't even do anything here because Hydra serves downloads by buffering them into memory already, which is terrible in all sorts of ways, I wish I hadn't looked into that to reply to this review comment :)

delft/hydra-proxy.nix
delroth: 2 requests in the last 14 days, not worth caring about.

delft/hydra-proxy.nix
delroth: Oh, and /nar/ isn't even a thing. (anymore?)

```
# zgrep /nar/ access-hydra.nixos.org.log.[23456].gz
[nothing]
```

mweinelt (approved):
Looks good to me. We can iterate further, once anything comes up.

delft/hydra-proxy.nix
Artturin: > Is there still a chromium wiki page that looks to a rather expensive hydra search which causes lots of traffic?

The page was switched to use search.nixos.org
https://chromium.googlesource.com/chromium/src/+/refs/heads/main/docs/linux/chromium_packages.md

PR #336, opened 2024-01-25
matrix-synapse: init on caliban

mweinelt:
Draft of the matrix-synapse deployment.

Does not yet build, because I'm waiting on UDS support for the synapse listeners, which people are currently actively working on.

- Generic nginx module
- Generic PostgreSQL module with backup integration
- Matrix Synapse module
- Promtheus scraping for Synapse

Depends on
- #345
- https://github.com/NixOS/nixpkgs/pull/286172

Follow-up:
- https://github.com/NixOS/nixos-homepage/pull/1218

### DIFF
--- a/non-critical-infra/modules/backup.nix
+++ b/non-critical-infra/modules/backup.nix
@@ -1,10 +1,24 @@
 { lib
 , config
+, pkgs
 , ...
 }:
 
 let
   cfg = config.services.backup;
+
+  mkZfsPreHook = mountpoint: ''
+    DATASET="$(findmnt -nr -o source "${mountpoint}")"
+    zfs snapshot -r "$DATASET@borg"
+
+    # https://github.com/borgbackup/borg/issues/6652
+    ls ${mountpoint}/.zfs/snapshot/borg/ > /dev/null
+  '';
+
+  mkZfsPostHook = mountpoint: ''
+    DATASET="$(findmnt -nr -o source "${mountpoint}")"
+    zfs destroy -r "$DATASET@borg"
+  '';
 in
 {
   options.services.backup = with lib; with types; {
@@ -71,6 +85,13 @@ in
         Paths to include in the backup.
       '';
     };
+    includesZfsDatasets = mkOption {
+      type = listOf str;
+      default = [];
+      description = ''
+        ZFS datasets referenced by mountpoint to snapshot and include
+      '';
+    };
 
     excludes = mkOption {
       type = listOf path;
@@ -105,7 +126,7 @@ in
     };
   };
 
-  config = lib.mkIf (cfg.includes != []) {
+  config = lib.mkIf (cfg.includes != [] || cfg.includesZfsDatasets != []) {
     programs.ssh.knownHosts."${if cfg.port != 22 then "[${cfg.host}]:${cfg.port}" else cfg.host}" = {
       publicKey = "${cfg.hostPublicKey}";
     };
@@ -113,6 +134,11 @@ in
     systemd.services.borgbackup-job-state = {
       wants = cfg.wantedUnits;
       after = cfg.wantedUnits;
+
+      path = lib.optionals (cfg.includesZfsDatasets != []) [
+        config.boot.zfs.package
+        pkgs.util-linux
+      ];
     };
 
     systemd.timers.borgbackup-job-state.timerConfig = {
@@ -122,7 +148,8 @@ in
     };
 
     services.borgbackup.jobs.state = {
-      inherit (cfg) preHook postHook;
+      preHook = lib.concatMapStringsSep "\n" mkZfsPreHook cfg.includesZfsDatasets;
+      postHook = lib.concatMapStringsSep "\n" mkZfsPostHook cfg.includesZfsDatasets;
 
       # Create the repo
       doInit = true;
@@ -136,11 +163,12 @@ in
       };
 
       # What to backup
-      paths = cfg.includes;
+      paths = cfg.includes ++
+        (map (mp: "${mp}/.zfs/snapshot/borg") cfg.includesZfsDatasets);
       exclude = cfg.excludes;
 
       # Where to backup it to
-      repo = "${cfg.user}@${cfg.host}:${config.networking.fqdn}/repo";
+      repo = "${cfg.user}@${cfg.host}:${config.networking.fqdn}";
       environment.BORG_RSH = "ssh -p ${cfg.port} -i ${cfg.sshKey}";
 
       # Ensure we don't fill up the destination disk
--- a/non-critical-infra/modules/matrix-synapse.nix
+++ b/non-critical-infra/modules/matrix-synapse.nix
@@ -1,0 +1,171 @@
+{ config
+, pkgs
+, ...
+}:
+
+{
+  imports = [
+    ./nginx.nix
+    ./postgresql.nix
+  ];
+
+  fileSystems."/var/lib/matrix-synapse" = {
+    device = "zroot/root/matrix-synapse";
+    fsType = "zfs";
+    options = [
+      "zfsutil"
+    ];
+  };
+
+  services.postgresql = {
+    ensureUsers = [ {
+      name = "matrix-synapse";
+      ensureDBOwnership = true;
+    } ];
+    # Insufficient to create the database with the correct collation
+    # https://github.com/element-hq/synapse/blob/develop/docs/postgres.md#set-up-database
+    ensureDatabases = [
+      "matrix-synapse"
+    ];
+  };
+
+  services.postgresqlBackup.databases = [
+    "matrix-synapse"
+  ];
+
+  services.redis.servers.matrix-synapse = {
+    enable = true;
+  };
+
+  environment.systemPackages = with pkgs; [
+    matrix-synapse-tools.synadm
+  ];
+
+  services.backup.includesZfsDatasets = [
+    "/var/lib/matrix-synapse"
+  ];
+
+  sops.secrets.matrix-synapse-signing-key = {
+    sopsFile = ../secrets/matrix-synapse-signing-key.caliban;
+    format = "binary";
+    path = "/var/lib/matrix-synapse/nixos.org.signing.key";
+    mode = "0600";
+    owner = "matrix-synapse";
+    group = "matrix-synapse";
+  };
+
+  sops.secrets.matrix-synapse-secrets = {
+    sopsFile = ../secrets/matrix-synapse-secrets.caliban;
+    format = "binary";
+    path = "/var/keys/matrix-synapse-secrets.conf";
+    mode = "0600";
+    owner = "matrix-synapse";
+    group = "matrix-synapse";
+  };
+
+  systemd.services.matrix-synapse.serviceConfig.SupplementaryGroups = [
+    "redis-matrix-synapse"
+  ];
+
+  #systemd.services.matrix-synapse.enable = false;
+
+  services.matrix-synapse = {
+    enable = true;
+    withJemalloc = true;
+
+    extraConfigFiles = [
+      config.sops.secrets.matrix-synapse-secrets.path
+    ];
+
+    # https://github.com/element-hq/synapse/blob/master/docs/usage/configuration/config_documentation.md
+    settings = {
+      enable_metrics = true;
+
+      server_name = "nixos.org";
+      signing_key_path = config.sops.secrets.matrix-synapse-signing-key.path;
+      public_baseurl = "https://matrix.nixos.org";
+      admin_contact = "infra@nixos.org";
+      web_client_location = "https://matrix.to/#/#community:nixos.org";
+
+      allow_public_rooms_over_federation = true;
+      allow_public_rooms_without_auth = true;
+
+      max_upload_size = "50M";
+
+      media_retention = {
+        local_media_lifetime = "90d";
+        remote_media_lifetime = "14d";
+      };
+
+      database = {
+        name = "psycopg2";
+        args = {
+          host = "/run/postgresql";
+        };
+      };
+
+      redis = {
+        enabled = true;
+        path = config.services.redis.servers.matrix-synapse.unixSocket;
+      };
+
+      listeners = [ {
+        type = "http";
+        # TODO: migrate to UNIX domain socket
+        #path = "/run/matrix-synapse/matrix-synapse.sock";
+        #mode = "0660";
+        port = 8008;
+        tls = false;
+        resources = [ {
+          compress = true;
+          names = [
+            "client"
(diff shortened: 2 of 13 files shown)


non-critical-infra/modules/matrix-synapse.nix
```diff
@@ -0,0 +1,112 @@
...
+        name = "psycopg2";
+        args = {
+          host = "/run/postgresql";
+        };
+      };
+
+      redis = {
+        enabled = true;
+        path = config.services.redis.matrix-synapse.path;
+      };
```
mweinelt, line 66: I think redis is only needed for workers, so we could leave it out.

non-critical-infra/modules/matrix-synapse.nix
```diff
@@ -0,0 +1,112 @@
...
+          ];
+        } {
+          compress = false;
+          names = [
+            "federation"
+          ];
+        } ];
+      } {
+        type = "http";
+        port = 8090;
```
zimbatm, line 85:
```suggestion
        port = 8090;
        bind_addresses = [ "127.0.0.1" ];
```

non-critical-infra/modules/matrix-synapse.nix
```diff
@@ -0,0 +1,112 @@
...
+
+      redis = {
+        enabled = true;
+        path = config.services.redis.matrix-synapse.path;
+      };
+
+      listeners = [ {
+        type = "http";
+        path = "/run/matrix-synapse/matrix-synapse.sock";
+        mode = "0660";
```
zimbatm, line 71:
```suggestion
        mode = "0660";
        x_forwarded = true;
```
so we get the right IPs in the logs

non-critical-infra/modules/matrix-synapse.nix
```diff
@@ -0,0 +1,112 @@
...
+    clientMaxBodySize = config.services.matrix-synapse.settings.max_upload_size;
+    upstreams."matrix-synapse" = {
+      "unix:/run/matrix-synapse/matrix-synapse.sock" = {};
+    }; 
+    virtualHosts."matrix.nixos.org" = {
+      forceSSL = true;
+      enableACME = true;
+
+      locations."^(/_matrix|/_synapse/client)" = {
+        proxyPass = "http://matrix-synapse";
```
zimbatm, line 105:
```suggestion
        proxyPass = "http://matrix-synapse";
        proxyWebsockets = true;
```
I have this in my config as well

non-critical-infra/modules/matrix-synapse.nix
```diff
@@ -0,0 +1,112 @@
...
+    settings = {
+      server_name = "nixos.org";
+      public_base = "https://matrix.nixos.org";
+      admin_contact = "infra@nixos.org";
+
+      allow_public_rooms_over_federation = true;
+      allow_public_rooms_without_auth = true;
+
+      max_upload_size = "50M";
+
```
zimbatm, line 55: It might be required to enable `experimental_features.spaces_enabled = true`, but maybe it's stable now.

non-critical-infra/modules/matrix-synapse.nix
mweinelt: > The x_forwarded option defaults to true when using Unix sockets and can be omitted.

https://github.com/element-hq/synapse/blob/master/docs/usage/configuration/config_documentation.md#listeners

mweinelt, line 71: Our synapse module does not support UNIX sockets yet. Work is happening over in https://github.com/dali99/nixos-matrix-modules/pull/7.

non-critical-infra/modules/matrix-synapse.nix
mweinelt: > - Remove the experimental `spaces_enabled` flag. The spaces features are always available now.

Removed in 2021. https://github.com/element-hq/synapse/blob/232adfb95f9ec726c2c09f3329c098aa9060c1f7/docs/changelogs/CHANGES-2021.md?plain=1#L1428

non-critical-infra/modules/matrix-synapse.nix
mweinelt: > bind_addresses: a list of local addresses to listen on. The default is 'all local interfaces'.

https://github.com/element-hq/synapse/blob/master/docs/usage/configuration/config_documentation.md#listeners

non-critical-infra/modules/matrix-synapse.nix
mweinelt: Does not seem like that is needed. I also don't have it on mine.

https://matrix-org.github.io/synapse/latest/reverse_proxy.html#nginx

non-critical-infra/modules/matrix-synapse.nix
mweinelt: The default bind is on IPv4/IPv6 wildcard, so public.

https://github.com/element-hq/synapse/blob/v1.99.0/synapse/config/server.py#L57

I configured IPv4/IPv6 localhost instead.

non-critical-infra/modules/backup.nix
```diff
@@ -0,0 +1,145 @@
...
+        monthly = 3;
+      };
+
+      # What to backup
+      paths = cfg.includes;
+      exclude = cfg.excludes;
+
+      # Where to backup it to
+      repo = "${cfg.user}@${cfg.host}:${config.networking.fqdn}";
+      environment.BORG_RSH = "ssh -p ${cfg.port} -i ~/.ssh/id_ed25519";
```
mweinelt, line 128: I think we can use the local root ssh key and not manage it any further. The ssh key install also happened manually.

https://docs.hetzner.com/robot/storage-box/backup-space-ssh-keys

non-critical-infra/modules/backup.nix
```diff
@@ -0,0 +1,145 @@
...
+      exclude = cfg.excludes;
+
+      # Where to backup it to
+      repo = "${cfg.user}@${cfg.host}:${config.networking.fqdn}/repo";
+      environment.BORG_RSH = "ssh -p ${cfg.port} -i ~/.ssh/id_ed25519";
+
+      # Ensure we don't fill up the destination disk
+      inherit (cfg) quota;
+
+      # Authenticated, Unencrypted
```
delroth, line 133: IMO we might want to change that with a per-machine secret so we can share storage boxes between several machines/services. But we can punt that to later.

non-critical-infra/modules/backup.nix
delroth: Can we have this use an SSH private key provisioned by sops instead?

JulienMalka:
@mweinelt Thank you for the great work on the backup module, do you mind if I cherry pick it to its own PR that we can merge so I can finish https://github.com/NixOS/nixos-org-configurations/pull/287 ?

mweinelt:
I extracted the backup module and caliban's config into #345

non-critical-infra/modules/backup.nix
mweinelt: Done in #345.

non-critical-infra/modules/backup.nix
mweinelt: Done in #345.

non-critical-infra/modules/matrix-synapse.nix
mweinelt: And now in https://github.com/NixOS/nixpkgs/pull/286172.

non-critical-infra/modules/matrix-synapse.nix
```diff
@@ -0,0 +1,130 @@
...
+  services.matrix-synapse = {
+    enable = true;
+    withJemalloc = true;
+
+    # https://github.com/element-hq/synapse/blob/master/docs/usage/configuration/config_documentation.md
+    settings = {
+      enable_metrics = true;
+
+      server_name = "nixos.org";
+      public_base = "https://matrix.nixos.org";
```
Ma27, line 62: Checked the docs and this option doesn't seem to exist, only `public_baseurl`.

non-critical-infra/modules/matrix-synapse.nix
```diff
@@ -0,0 +1,130 @@
...
+  services.nginx = {
+    clientMaxBodySize = config.services.matrix-synapse.settings.max_upload_size;
+    upstreams."matrix-synapse" = {
+      "unix:/run/matrix-synapse/matrix-synapse.sock" = {};
+    }; 
+    virtualHosts."matrix.nixos.org" = {
+      forceSSL = true;
+      enableACME = true;
+
+      locations."^(/_matrix|/_synapse/client)" = {
```
Ma27, line 122: Don't you also need `/.well-known/matrix/server` to delegate the homeserver for `nixos.org` to `matrix.nixos.org`?

non-critical-infra/modules/matrix-synapse.nix
```diff
@@ -0,0 +1,130 @@
...
+          compress = false;
+          names = [
+            "federation"
+          ];
+        } ];
+      } {
+        type = "http";
+        bind_addresses = [
+          "127.0.0.1"
+          "::1"
```
Ma27, line 101: Oh, does the metrics listener not crash anymore when instructing it to listen on a v6 address?
I'm pretty sure that this was the case I only have it listening on v4.

non-critical-infra/modules/matrix-synapse.nix
mweinelt: Yeah, but that will happen over in nixos-homepage, since we're hosting nixos.org.

Ma27: Ah right, wasn't aware that this isn't part of the configs in here, disregard then.

non-critical-infra/modules/matrix-synapse.nix
mweinelt: Mine listens on IPv6 wildcard only and that works. I have not deployed this yet.

non-critical-infra/modules/matrix-synapse.nix
mweinelt: https://github.com/NixOS/nixos-homepage/pull/1218

non-critical-infra/modules/matrix-synapse.nix
```diff
@@ -0,0 +1,127 @@
...
+            "metrics"
+          ];
+        } ];
+      } ];
+    };
+  };
+
+  services.nginx = {
+    clientMaxBodySize = config.services.matrix-synapse.settings.max_upload_size;
+    upstreams."matrix-synapse" = {
```
Ma27, line 112: Isn't this supposed to be `upstreams.<name>.servers.<upstream>`?

non-critical-infra/modules/matrix-synapse.nix
Ma27: That's cool, just realized that prometheus_client supports that now, nice :tada: 
In other words, this can be resolved.

non-critical-infra/modules/matrix-synapse.nix
Ma27: How does nginx communicate with the socket btw? Perhaps I'm missing something, but it needs to be part of the matrix-synapse group for that, right?

mweinelt: Absoutely, passing the group via `SupplementaryGroups=` was the plan.

Erethon:
I would recommend adding some extra configs in synapse, namely:
* [federation_metrics_domains](https://matrix-org.github.io/synapse/latest/usage/configuration/config_documentation.html#federation_metrics_domains) pointing to `matrix.org` and maybe other big public instances (like `fairydust.space`). This can help with identifying problems like the ones mentioned in #410
* [web_client_location](https://matrix-org.github.io/synapse/latest/usage/configuration/config_documentation.html#web_client_location): This will help any people that end up browsing to `matrix.nixos.org` without really knowing what to do to join the chat. Since `matrix.nixos.org` won't actually host any user accounts it doesn't make sense to point this to an actual element-web client.  I think it could be set to https://matrix.to/#/#community:nixos.org which is the same link as the one provided in the [community page](https://nixos.org/community/).
* [presence](https://matrix-org.github.io/synapse/latest/usage/configuration/config_documentation.html#presence): Presence is a feature that's know to [cause issues](https://github.com/element-hq/synapse/issues/15877) and is often disabled in most large installations as mentioned in the documentation. Since this homeserver will only host bot accounts, there really isn't a reason to have presence enabled that is a feature meant for humans and is resource heavy.
* [media_retention](https://matrix-org.github.io/synapse/latest/usage/configuration/config_documentation.html#media_retention): This is more of a policy thing, but I believe that remote media should be removed after a while because they're somewhat of an abuse vector currently.

mweinelt:
> [federation_metrics_domains](https://matrix-org.github.io/synapse/latest/usage/configuration/config_documentation.html#federation_metrics_domains) pointing to matrix.org and maybe other big public instances (like fairydust.space). This can help with identifying problems like the ones mentioned in https://github.com/NixOS/infra/issues/410

We'll revisit that when we dig into monitoring.

mweinelt:
> [web_client_location](https://matrix-org.github.io/synapse/latest/usage/configuration/config_documentation.html#web_client_location): This will help any people that end up browsing to matrix.nixos.org without really knowing what to do to join the chat. Since matrix.nixos.org won't actually host any user accounts it doesn't make sense to point this to an actual element-web client. I think it could be set to [matrix.to/#/#community:nixos.org](https://matrix.to/#/#community:nixos.org) which is the same link as the one provided in the [community page](https://nixos.org/community/).

Implemented.

> [presence](https://matrix-org.github.io/synapse/latest/usage/configuration/config_documentation.html#presence): Presence is a feature that's know to https://github.com/element-hq/synapse/issues/15877 and is often disabled in most large installations as mentioned in the documentation. Since this homeserver will only host bot accounts, there really isn't a reason to have presence enabled that is a feature meant for humans and is resource heavy.

We have one interactive administrative user, and we're not that short on resources.

> [media_retention](https://matrix-org.github.io/synapse/latest/usage/configuration/config_documentation.html#media_retention): This is more of a policy thing, but I believe that remote media should be removed after a while because they're somewhat of an abuse vector currently.

Went with 90d for local content, 14d for remote content, as shown in the config example from upstream.

mweinelt:
We're migrated.

Set up the [Grafana dashboard](https://grafana.nixos.org/d/000000012/matrix-synapse?orgId=1&var-DS_PROMETHEUS=000000001&var-bucket_size=$__auto_interval_bucket_size&var-instance=matrix.nixos.org:443&var-job=All&var-index=All&from=now-30m&to=now) up and configured `federation_metrics_domains`.

PR #337, opened 2024-01-26
hydra: serve favicon

mweinelt:
Browsers request favicon.ico, so we might as well.


delroth:
Can we fix that in Hydra instead by adding that image to their /static/ and the right rel="icon" tag in the HTML templates? That sounds like a better solution to me. WDYT?

Issue #346, opened 2024-02-03
prometheus-nixos-exporter traces on rhea


mweinelt:
The following trace is visible every 15 seconds when Prometheus queries.

```pytb
Traceback (most recent call last):
  File "/nix/store/p1zbyfrpj3hq50mxh5hmxl3kqpa2b1am-python3-3.11.6/lib/python3.11/wsgiref/handlers.py", line 137, in run
    self.result = application(self.environ, self.start_response)
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/nix/store/57qzqalpdgk4p2rjys669wn8lkbw99mm-python3-3.11.6-env/lib/python3.11/site-packages/prometheus_client/exposition.py", line 129, in prometheus_app
    status, headers, output = _bake_output(registry, accept_header, accept_encoding_header, params, disable_compression)
                              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/nix/store/57qzqalpdgk4p2rjys669wn8lkbw99mm-python3-3.11.6-env/lib/python3.11/site-packages/prometheus_client/exposition.py", line 105, in _bake_output
    output = encoder(registry)
             ^^^^^^^^^^^^^^^^^
  File "/nix/store/57qzqalpdgk4p2rjys669wn8lkbw99mm-python3-3.11.6-env/lib/python3.11/site-packages/prometheus_client/openmetrics/exposition.py", line 21, in generate_latest
    for metric in registry.collect():
  File "/nix/store/57qzqalpdgk4p2rjys669wn8lkbw99mm-python3-3.11.6-env/lib/python3.11/site-packages/prometheus_client/registry.py", line 97, in collect
    yield from collector.collect()
  File "/nix/store/lvxc959d2rvl6qxwkzx72nfjcgsabh6x-nixos-exporter.py", line 21, in collect
    self.get_time("/run/current-system"))
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/nix/store/lvxc959d2rvl6qxwkzx72nfjcgsabh6x-nixos-exporter.py", line 50, in get_time
    return parsed[0]['registrationTime']
           ~~~~~~^^^
KeyError: 0
```

PR #347, opened 2024-02-03
prometheus-nixos-exporter: add nix 2.19.0 compat

mweinelt:
The output of `nix path-info --json` changed from a list to an object in nix 2.19.0.

Add a simple version detection, so we can support both output formats.

Fixes: #346

### DIFF
--- a/modules/prometheus/nixos-exporter.py
+++ b/modules/prometheus/nixos-exporter.py
@@ -33,24 +33,29 @@ class NixosSystemCollector:
         current_system = GaugeMetricFamily(
             "nixos_current_system_time_seconds",
             "The time the system's current generation was registered in the Nix database.",
-            labels=["version_id"]
+            labels=["version_id"],
         )
-        current_system.add_metric([self.get_version_id("/run/current-system")],
-                                  self.get_time("/run/current-system"))
+        current_system.add_metric(
+            [self.get_version_id("/run/current-system")],
+            self.get_time("/run/current-system"),
+        )
         yield current_system
 
         booted_system = GaugeMetricFamily(
             "nixos_booted_system_time_seconds",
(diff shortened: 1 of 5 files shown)


modules/prometheus/nixos-exporter.py
```diff
@@ -40,14 +58,20 @@ def get_version_id(self, path):
...
         result = subprocess.run(
             [ "nix", "path-info", "--json", path ],
             stdout=subprocess.PIPE
         )
         if result.returncode == 0:
             parsed = json.loads(result.stdout)
-            return parsed[0]['registrationTime']
+
+            # https://github.com/NixOS/nix/pull/9242
+            if self.nix_version >= "2.19.0":
```
vcunat, line 70: Will that work?
```
$ python3
>>> "2.3.16" >= "2.19.0"
True
```

vcunat: This one perhaps?
https://stackoverflow.com/a/11887825/587396

vcunat: I don't think we need wide version compatibility here, but the linked approach seems easy enough.

vcunat:
Overall I think this is OK to just deploy and see.

modules/prometheus/nixos-exporter.py
mweinelt: Ah, yeah. That is annoying!

mweinelt: Replaced the version check with a pretty standard library and tested both 2.18.1 and 2.19.0.

mweinelt:
> Overall I think this is OK to just deploy and see.

Agreed, I just need to bootstrap the VPN on my desktop.

Issue #324, opened 2024-01-06
Rework access management & deployment for the NixOS core infra


delroth:
Please take this as an RFC and feel free to yell at what seems to be a bad idea and/or suggest improvements.

### Goals
- Unify access management, currently very fragmented.
  - Some hosts only accessible via `nixops ssh` from `bastion`, some only accessible via `ssh` on Wireguard, varying login usernames depending on the host.
- Unify deployment mechanisms.
  - Currently a mix of `nixops` and manual on-host `nixos-rebuild`.
- Introduce proper secret management.
  - Currently a mix of unversioned files in a directory on `bastion` deployed via `nixops`, and manually deployed secret files on `rhea` and `bastion` .
- Only change as much as needs to be changed for meet those goals. No scope creep.

-----

### Plan
- More common Nix configuration for users and access management on core infra machines.
  - Get rid of automagic NixOps SSH keys, have `root` SSH access for all core infra SSH keys.
  - Potentially: switch to separate users with sudo access for SSH administrative access - it doesn't grant us much in terms of security, but it has some auditability benefits.
- Manage secrets with `sops-nix`.
  - The most critical secrets will be kept out of scope for this out of caution (for now). Specifically identified now: the Hydra signing key. A full inventory will be necessary to identify any other critical secrets. Rule of thumb: if it can't be rotated, let's keep managing it "manually" for now.
  - This aligns with the infra choice already made for `non-critical-infra`.
- Deprecate `nixops` completely.
  - We already don't use any of its advanced features, it's really just deploying configs from a flake.
  - We'll switch to a flake with a `outputs.nixosConfigurations` per machine, as well as `outputs.colmena` for `colmena` compat for remote deployment.
  - There is a total of 2 use cases of listing all machines in the nixops network, for monitoring, and that's kinda broken already because that doesn't include `rhea`. So for now, until new use cases appear for this, let's stop relying on any cross-machine configuration and keep every `nixosConfiguration` independently evaluable.

-----

## Future improvements

- Move channel scripts out of `bastion` and get rid of it.
  - $120/month savings, and once `nixops` is dead + secrets versioned, the use case for `bastion` kind of goes away (no more unversioned state required for a deployment tool).
  - It doesn't seem to me that channel scripts currently need to be colocated with the cache, but I haven't looked in too much details yet.
- Figure out what we want to do with Wireguard
[...]

PR #350, opened 2024-02-08
Remove NixOps requirement for eris/haumea

delroth:
This PR removes the requirement to use NixOps for eris/haumea deployment. Specifically:
- It fixes some impure dependencies to local files on `bastion`, incl. Nix expressions that were previous unversioned.
- It imports into the repository some state that was implicitly maintained by NixOps.
- It replaces NixOps secrets management with agenix.
- It fixes and simplifies a few things regarding `stateVersion` and `delft/common.nix` that were confusing and caused issues during the port.

This has been successfully test-deployed with:
```
$ nixos-rebuild --flake .#eris --target-host root@eris.nixos.org --use-substitutes test
$ nixos-rebuild --flake .#haumea --target-host root@haumea.nixos.org --use-substitutes test
```

Ref #324.

There are a ton of things left to clean up here, but most of the ugliness is ugliness that already existed unversioned on `bastion` before, and it just makes it explicit. We can start doing extensive cleanups faster now that we don't need to use `bastion` anymore to build/test the NixOS configs for those machines.


delft/secrets.nix
```diff
@@ -0,0 +1,19 @@
+let
+  keys = import ../ssh-keys.nix;
+
+  secrets = with keys; {
+    alertmanager-matrix-forwarder = [ machines.eris ];
+    fastly-read-only-api-token = [ machines.eris ];
+    packet-sd-env = [ machines.eris ];
+    prometheus-packet-spot-market-price-exporter = [ machines.eris ];
+    rfc39-credentials = [ machines.eris ];
+    rfc39-github = [ machines.eris ];
```
mweinelt, line 10: Decrypting that leads to a binary file, and I don't see where this secret gets used.

mweinelt: Referenced in `rfc39-credentials` :shrug:

mweinelt (approved):
Went through all commits. Looks like progress to me!

PR #357, opened 2024-02-11
Remove live.nixos.org from DNS

bryanhonof:
Had to pull the server down, since I don't have time to maintain it anymore.


JulienMalka:
Thank you. We can revive this on the non-critical infra later if there is a need for that service.

mweinelt:
DNS changes deployed.

delroth:
@JulienMalka in the future can we wait until someone who can deploy the changes has looked at them before they get merged? Otherwise there's the potential for these to get missed/lost.

PR #361, opened 2024-02-11
delft: update network configuration

mweinelt:
- Migrates network configuration to networkd 
- Drops udev rules for network interface names, returns to predictable interface names
  - Should be impact-free, since we don't match the interface name anyway
- Migrates DNS configuration to resolved
- Moves eris' IPv6 address from the network address
- Sets up IPv6 DNS for eris

### DIFF
--- a/delft/eris/network.nix
+++ b/delft/eris/network.nix
@@ -1,0 +1,24 @@
+{
+  systemd.network = {
+    enable = true;
+    networks = {
+      "30-enp0s31f6" = {
+        matchConfig = {
+          MACAddress = "90:1b:0e:91:c3:67";
+          Type = "ether";
+        };
+        linkConfig.RequiredForOnline = true;
+        networkConfig.Description = "WAN";
+        address = [
+          "138.201.32.77/26"
+          "2a01:4f8:171:33cc::1/64"
+        ];
+        routes = [ {
+          routeConfig.Gateway = "138.201.32.65";
+        } {
+          routeConfig.Gateway = "fe80::1";
+        } ];
+      };
+    };
+  };
+}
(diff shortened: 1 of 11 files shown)


mweinelt:
Deployed with test and switch. Did not notice any impact.

delroth:
Shall we try to do some reboots? Sounds exciting.

mweinelt:
We really should. The devil is in the details.

delft/rhea/network.nix
```diff
@@ -0,0 +1,34 @@
+{
+  networking.hostId = "9cd372da";
+
+  services.resolved = {
+    enable = true;
+    fallbackDns = [
+      "185.12.64.1"
```
delroth, line 7: I really see no reason to have this list be different from the others btw - let's fix it along the way maybe? Before we forget :p

delft/eris/network.nix
```diff
@@ -0,0 +1,36 @@
...
+        linkConfig.RequiredForOnline = true;
+        networkConfig.Description = "WAN";
+        address = [
+          "138.201.32.77/26"
+          "2a01:4f8:171:33cc::1/64"
+        ];
+        routes = [ {
+          routeConfig.Gateway = "138.201.32.65";
+        } {
+          routeConfig.Gateway = "fe80::1";
```
delroth, line 31: FWIW on my Hetzner machines I used to need GatewayOnLink=true for this to work. Not sure if that's still the case, but worth keeping in mind if we run into issues with v6 connectivity.

delroth (approved):

delft/eris/network.nix
mweinelt: We'll have link local addresses, so it will he in the subnet, no onlink flag required in that case. 

https://www.freedesktop.org/software/systemd/man/latest/systemd.network.html#LinkLocalAddressing=

PR #372, opened 2024-02-13
eris: alert on failed systemd services

delroth:
There might be some initial flakiness due to some badly behaved services, e.g. hydra-scale-equinix-metal, but nothing that we can't silence away and properly fix later.

15m is a very conservative threshold, I'm expecting to push it down at some point once things get more under control.


delroth:
Pushed for testing: https://monitoring.nixos.org/prometheus/alerts?search=systemd

mweinelt (approved):

PR #373, opened 2024-02-13
Migrate nix configuration into nix.settings

mweinelt:
This removes the deprecation warnings from the nix module. The sandbox config is dropped on purpose, since the sandbox is enabled by default.


zimbatm (approved):

mweinelt:
Deployed on eris, haumea and rhea.

PR #374, opened 2024-02-13
Add CI and pre-commit hooks for nix files

mweinelt:


Mic92:
Looks like we have CI now more or less?

PR #378, opened 2024-02-16
prometheus-nixos-exporter: Expose whether current-system kernel is booted

mweinelt:
This allows alerting on when the current-system has a different kernel, than the booted one, which usually indicates we should reboot the machine.

@maralorn originally wanted this, and I proposed we extend the exporter for the benefit of NixOS infra.

```
nixos_current_system_kernel_booted{booted="/nix/store/gp3pl6swvksnj9ws646fnlz553w079pb-linux-6.1.77", current="/nix/store/gp3pl6swvksnj9ws646fnlz553w079pb-linux-6.1.77", instance="eris:9300", job="nixos", role="monitoring"} 1

### DIFF
--- a/modules/prometheus/nixos-exporter/prometheus_nixos_exporter/__main__.py
+++ b/modules/prometheus/nixos-exporter/prometheus_nixos_exporter/__main__.py
@@ -4,6 +4,7 @@
 
 import subprocess
 import json
+import os
 import sys
 import time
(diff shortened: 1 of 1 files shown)


delroth (approved):

mweinelt:
We'll talk about the prometheus rule next, we're not in a hurry anyway.

Issue #370, opened 2024-02-12
Clarify existing licensing and apply MIT license


mweinelt:
The repo is currently published without a license, which is not ideal for reusability or for porting things into nixpkgs, which requires contributions to be MIT licensed.

I think going for MIT would be easiest, but I'm open to discussing alternatives. Also, we would need the sign-off of every previous contributor, which feels like a daunting task.

* [x] @edolstra
* [x] @grahamc
* [x] @zimbatm
* [x] @delroth
* [x] @rbvermaa
* [x] @mweinelt
* [x] @cole-h
* [x] @lheckemann
* [x] @garbas
* [x] @vcunat
* [x] @JulienMalka
* [x] @domenkozar
* [x] @LnL7

edolstra:
Probably not worth it, given that this is an "internal" repo. But we could ask that any new contributions are MIT-licensed.

delroth:
Very similar precedent: https://github.com/NixOS/nixos-search/pull/723

Let's start by figuring out how many contributors we'd need to get an approval from to relicense. I agree that MIT licensing makes the most sense to align with every other official NixOS-related project. I'm reasonably happy with using the same guideline/threshold as nixos-search, aka. "more than 100 lines contributed".

vcunat:
Only 9 of them over 100 lines I think: https://github.com/NixOS/infra/graphs/contributors

But I'm no lawyer; not sure how safe it is to use such a simple metric.

zimbatm:
I hereby agree to release all my changes under the MIT :+1: 

(also created the list above)

JulienMalka:
I hereby agree to release all my changes under the MIT 👍

rbvermaa:
I hereby agree to release all my changes under the MIT 👍

delroth:
I hereby agree to release all my changes under the MIT 👍

vcunat:
I hereby agree to release all my changes under the MIT 👍

domenkozar:
I hereby agree to release all my changes under the MIT 👍

garbas:
I hereby agree to release all my changes under the MIT 👍

mweinelt:
I hereby agree to release all my changes under the MIT 👍

lheckemann:
I hereby agree to release all my changes under the MIT 👍

cole-h:
I hereby agree to release all my changes under the MIT 👍

grahamc:
MIT seems good!

LnL7:
[...]

PR #380, opened 2024-02-20
Create LICENSE

zimbatm:
Optimistic lock on #370

Based on my knowledge of the people involved, it might take a while, but I don't think anybody will be against licensing under MIT.

Meanwhile, this allows the clarification of new contributions under the license.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,21 @@
+MIT License
+
+Copyright (c) 2024 NixOS Foundation and contributors
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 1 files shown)


delroth:
> Meanwhile, this allows the clarification of new contributions under the license.

Unfortunately it also implies that the contents of the repo are MIT licensed, which they currently aren't (at least not completely). So NAK on merging this before we have collected approvals.

I get your intention though - but I don't think it's necessary. People with direct access to the repo have all already approved or are already in the list, and in the eventuality of a PR from a new contributor we can ask them in the PR.

zimbatm:
Everybody approved now

JulienMalka (approved):

PR #385, opened 2024-02-21
pluto: init

mweinelt:
The machine is live. DNS is updated.

### DIFF
--- a/delft/pluto/disko.nix
+++ b/delft/pluto/disko.nix
@@ -1,0 +1,90 @@
+{
+  disko.devices = {
+    disk = {
+      nvme0n1 = {
+        type = "disk";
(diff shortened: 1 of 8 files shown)


delft/pluto/disko.nix
```diff
@@ -0,0 +1,89 @@
...
+    };
+    zpool = {
+      zroot = {
+        type = "zpool";
+        options = {
+          ashift = "12";
+        };
+        mode = "mirror";
+        rootFsOptions = {
+          acltype = "posixacl";
```
delroth, line 75: Could you add autotrim in there as well? And apply it to the live server.

delroth (approved):

delft/pluto/disko.nix
mweinelt: Added and redeployed.

```
# zpool get autotrim
NAME   PROPERTY  VALUE     SOURCE
zroot  autotrim  on        local
```

PR #390, opened 2024-02-24
delft: cleanup, eris config split

mweinelt:
Deployed to eris, prometheus targets results look the same.


delroth (approved):
I did a quick pass over everything, it looks fine. If it evals and does the right thing let's go with it :)

PR #399, opened 2024-03-06
add SPF/DMARC/DKIM for wiki.nixos.org

Mic92:


mweinelt (approved):

terraform/dns.tf
```diff
@@ -291,6 +291,21 @@ locals {
       type     = "A"
       value    = "65.21.240.250"
     },
+    {
+      hostname = "wiki.nixos.org"
+      type     = "TXT"
+      value    = "v=spf1 ip4:65.21.240.250 ip6:2a01:4f9:c012:8178:: ~all"
```
mweinelt, line 297: It is uncommon to use the network address, but I'm not aware of the downsides right now.

zimbatm: yeah, we still need to test this a bit. I'll send another PR if this needs to be fixed.

mweinelt: Would generally advise to just use `::1`.

Issue #401, opened 2024-03-13
Meeting infrastructure


Janik-Haag:
**Is your feature request related to a problem? Please describe.**

Currently we rely a lot on @lassulus meeting infra (mostly pad and jitsi), it would be awesome to have a official alternative that is a better fit. This would probably be a good fit for non-critical infra

cc @JulienMalka

mweinelt:
One year later and I would suggest going for https://github.com/suitenumerique/meet instead.

fricklerhandwerk:
Why not Jitsi?

mweinelt:
I tested meet, and it felt much more sleek and professional, promises to scale better and uses livekit instead of a bespoke stack that includes an XMPP server.

PR #404, opened 2024-03-22
non-critical-infra: init jitsi

Janik-Haag:
***This config still contains a bug where lobby-autostart kicks the moderator after joining using the secure domain bypass***

The jitsi config has the lobby enabled for every-room by-default, this means only people that get access will be able to join. Access control is done using the `Secure Domain Lobby Bypass` plugin and should be changed to `Token Lobby Bypass` once we have a IDP.

Adding users for Secure Domain Lobby Bypass has to be done in a imperativ way using:
```bash
prosodyctl register <username> jitsi.nixos.org <password>
```

Depends on [nixos/nixpkgs#297809](https://github.com/NixOS/nixpkgs/pull/297809)
And the DNS records are still missing from the pr.

### DIFF
--- a/non-critical-infra/hosts/caliban.nixos.org/default.nix
+++ b/non-critical-infra/hosts/caliban.nixos.org/default.nix
@@ -9,6 +9,7 @@
       ../../modules/first-time-contribution-tagger.nix
       ../../modules/backup.nix
       ../../modules/vaultwarden.nix
+      ../../modules/jitsi.nix
     ];
 
   # Bootloader.
--- a/non-critical-infra/modules/jitsi.nix
+++ b/non-critical-infra/modules/jitsi.nix
@@ -1,0 +1,59 @@
+{ pkgs, ... }:
+{
+  services.jitsi-meet = {
+    enable = true;
+    hostName = "jitsi.nixos.org";
+    config = {
+      enableWelcomePage = true;
+      requireDisplayName = true;
+      analytics.disabled = true;
+      startAudioOnly = true;
+      channelLastN = 4;
+      lobby = {
+        autoKnock = true;
+        enableChat = false;
+      };
+      stunServers = [
+        { urls = "turn:turn.matrix.org:3478?transport=udp"; }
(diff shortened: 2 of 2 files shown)


delroth:
ref #401

delroth:
> And the DNS records are still missing from the pr.

Can you maybe send that as a separate PR ahead of time? Do we know on which machine we're deploying this at this point? (caliban?)

non-critical-infra/hosts/caliban.nixos.org/default.nix
```diff
@@ -9,6 +9,7 @@
       ../../modules/first-time-contribution-tagger.nix
       ../../modules/backup.nix
       ../../modules/vaultwarden.nix
+      ../../modules/jitsi.nix
```
delroth, line 12: Please keep this sorted.

delroth: Oh, looks like it's already not. Bleh.

So instead: can you sort this? :P

non-critical-infra/modules/jitsi.nix
```diff
@@ -0,0 +1,59 @@
...
+      enableWelcomePage = true;
+      requireDisplayName = true;
+      analytics.disabled = true;
+      startAudioOnly = true;
+      channelLastN = 4;
+      lobby = {
+        autoKnock = true;
+        enableChat = false;
+      };
+      stunServers = [
```
Erethon, line 16: This should be `p2p.stunServers` according to the [documentation](https://jitsi.github.io/handbook/docs/devops-guide/turn) and [code](https://github.com/jitsi/jitsi-meet/blob/cc03949b90121350648503c6d2f8a8041f7245cc/config.js#L1005).

An alternative to using (and depending on) the matrix.org TURN server would be to disable the p2p calls for two person calls completely with `p2p.enabled = false`. One could argue this is a bit more privacy friendly, since this way you never expose your IP to the other person on the call.

mweinelt:
Stale.

Issue #212, opened 2022-07-04
cache.nixos.org: fastly<->s3 throttled? 


srhb:
**Affected service**
cache.nixos.org

**Describe the issue**
For, I think, a few weeks now, I've noticed very poor download speeds (~1-2MB/s max) on what I will describe as "fresh paths" in the nixos cache -- paths that I expect have not recently been fetched by anyone else (in my region?)
After a download, an immediate redownload of the same path is extremely fast, like it used to be (~100MB/s). This makes me suspect the problem is "behind" the Fastly cache, which I assume is s3.

My test to verify it wasn't a nix issue was a direct wget (below) of openjdk11, but because of the nature of the problem, you may have to find another, arbitrary hugepath to test with that hasn't been warmed recently (in your region?)

```
wget https://cache.nixos.org/nar/1bd1ji8wghy39rnk5xsb56vw81nbq961rcm42lxid5j4iga9pivc.nar.xz
```

SuperSandro2000:
Noticed the same lately

mweinelt:
The firefox-unwrapped package is well cached for a comparison:

```
wget https://cache.nixos.org/nar/17vh4y9w3xnwibw6a11bbpjd8zhdfkfndvjl3g8bx92pwrbsd25h.nar.xz
```

Getting 4-5 MB/s uncached and 24 MB/s cached.

cdepillabout:
This was separately reported on Discourse as well: https://discourse.nixos.org/t/slow-cache-nixos-org/20131

domenkozar:
cc @thoughtpolice

edolstra:
Maybe https://github.com/NixOS/nixos-org-configurations/commit/5e0fb0146b6ba4f053d452c1c3e0e28ac06ee443 had some unintended performance consequences? That was on June 23rd.

vcunat:
On a quick glance, [Shielding](https://docs.fastly.com/en/guides/shielding) could also increase traffic to S3 significantly, and June did see an increase in the totals.

edolstra:
My understanding is that it only increases traffic in the case where two users start downloading a file at the same time, which is probably not extremely common.

wahjava:
the speeds are still terrible:

https://asciinema.org/a/u9gzwyK1JG1rh9fnwDiLymsZ6

Could someone maybe file a ticket with Fastly, it's possible they may have some network congestion in path to origin (S3?), my 0.02 XDR ?

Thanks in advance

[...]

PR #409, opened 2024-04-10
fastly: use shielding POPs that match backend location

delroth:
We were previously always shielding via the Virginia POP which is closest to us-east-1. However the S3 buckets for releases.nixos.org and tarballs.nixos.org are hosted in eu-west-1 (Dublin). Let's save on some transatlantic roundtrips, and maybe even improve #212.

### DIFF
--- a/terraform/locals.tf
+++ b/terraform/locals.tf
@@ -4,8 +4,6 @@ locals {
   # TLS v1.2, protocols HTTP/1.1 and HTTP/2
   fastly_tls12_sni_configuration_id = "5PXBTa6c01Xoh54ylNwmVA"
 
-  fastly_shield = "iad-va-us"
-
   cache-iam  = data.terraform_remote_state.terraform-iam.outputs.cache
   fastlylogs = data.terraform_remote_state.terraform-iam.outputs.fastlylogs
 
--- a/terraform/cache.tf
+++ b/terraform/cache.tf
@@ -122,7 +122,7 @@ resource "fastly_service_vcl" "cache" {
     name                  = "s3.amazonaws.com"
     override_host         = aws_s3_bucket.cache.bucket_domain_name
     port                  = 443
-    shield                = local.fastly_shield
+    shield                = "iad-va-us"
     ssl_cert_hostname     = "s3.amazonaws.com"
     ssl_check_cert        = true
     use_ssl               = true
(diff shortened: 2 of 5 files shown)


delroth:
```
  # fastly_service_vcl.nixpkgs-tarballs will be updated in-place
  ~ resource "fastly_service_vcl" "nixpkgs-tarballs" {
      ~ active_version     = 29 -> (known after apply)
      ~ cloned_version     = 29 -> (known after apply)
        id                 = "3z4j8BCi1YqXEqGuwkFn6Q"
        name               = "tarballs.nixos.org"
        # (9 unchanged attributes hidden)

      - backend {
          # At least one attribute in this block is (or was) sensitive,
          # so its contents will not be displayed.
        }
      + backend {
          # At least one attribute in this block is (or was) sensitive,
          # so its contents will not be displayed.
        }

        # (12 unchanged blocks hidden)
    }

  # fastly_service_vcl.releases will be updated in-place
  ~ resource "fastly_service_vcl" "releases" {
      ~ active_version     = 53 -> (known after apply)
      ~ cloned_version     = 53 -> (known after apply)
        id                 = "745rWITBeKgJklfVqQ2Sjb"
        name               = "releases.nixos.org"
        # (8 unchanged attributes hidden)

      - backend {
          # At least one attribute in this block is (or was) sensitive,
          # so its contents will not be displayed.
        }
      + backend {
          # At least one attribute in this block is (or was) sensitive,
          # so its contents will not be displayed.
        }

        # (13 unchanged blocks hidden)
    }

Plan: 0 to add, 2 to change, 0 to destroy.
```

terraform/locals.tf
```diff
@@ -4,8 +4,6 @@ locals {
   # TLS v1.2, protocols HTTP/1.1 and HTTP/2
   fastly_tls12_sni_configuration_id = "5PXBTa6c01Xoh54ylNwmVA"
 
-  fastly_shield = "iad-va-us"
```
cole-h, line 7: Maybe we could have like a "dict" (I dunno what Terraform calls it) for the shields?

I'm imagining something like (maybe invalid syntax but hopefully you get the idea):

```terraform
fastly_shield = {
  cache = "iad-va-us"
  channels = "iad-va-us"
  tarballs = "dub-dublin-ie"
  releases = "dub-dublin-ie"  
}
```

delroth: I'm not particularly opposed to it but I also don't see the benefit compared to just inlining it in the Fastly service configuration here. There's plenty of service-specific knobs that we're not configuring through such maps right now.

mweinelt (approved):

delroth:
```
Apply complete! Resources: 0 added, 2 changed, 0 destroyed.
```

image

PR #415, opened 2024-04-16
fastly: Redirect http to https for some websites

Erethon:
Opening this after some [discussion in the Matrix Infra channel](https://matrix.to/#/!RROtHmAaQIkiJzJZZE:nixos.org/$bsxoT7OErMpVDSGfLPIAOdzexbQ1jF5ltKJyOsrSVvc?via=nixos.org&via=matrix.org&via=nixos.dev).

(cache|channels|tarballs|releases).nixos.org currently don't redirect http to https and are available under http. I believe the changes in this PR will make Fastly redirect http to https, but I have no way to test this, this assumption is based on the docs of the Terraform provider.


delroth (approved):
Dumping the Matrix discussion here for future reference:

```
<vcunat>
 These are not websites, really.
<vcunat> Off the top of my head I'm not 100% about possible use cases for plain http in some of these. Say, for [cache.nixos.org](http://cache.nixos.org/) the content is signed and public, so the value of https is a bit lower than usual.
<delroth> there's no good reason to serve HTTP without TLS in 2024
<delroth> if someone has a use case for it, it's too bad for them
<vcunat> Say, transparent caching proxies?
<vcunat> I've heard about a big one in China, but no idea how it's set up.  (and there might be more)
<vcunat> (Just a bit of brainstorming, nothing serious on my side.)
<delroth> transparent caching proxies wouldn't work with default nix/nixos configuration which uses https://cache.nixos.org/ as the substituter, so you might as well tell your users to use https://my.cache.foo.com/ instead of telling them to downgrade to http://cache.nixos.org/ so you can MITM their traffic
<delroth> [releases.nixos.org](http://releases.nixos.org/) not redirecting to HTTPS and not using HSTS is a major bug imo
<delroth> because that's not verified by the user agent in most situations
<delroth> (there's no signature, the sha256 is stored on the same origin, and nobody checks the sha256 anyway)
```

delroth:
Deployed and seems to be working. Thank you for sending this patch!

```
$ curl -I http://cache.nixos.org/
HTTP/1.1 301 Moved Permanently
Connection: close
Content-Length: 0
Server: Varnish
Retry-After: 0
Location: https://cache.nixos.org/
Accept-Ranges: bytes
Date: Tue, 16 Apr 2024 19:23:05 GMT
Via: 1.1 varnish
X-Served-By: cache-vie6342-VIE
X-Cache: HIT
X-Cache-Hits: 0
X-Timer: S1713295386.891547,VS0,VE0
access-control-allow-origin: *
```

delroth:
Next step is probably making sure we have HSTS headers enabled, but that's a more dangerous step since it can't easily be rolled back. Let's revisit in a few months if nobody has yelled for this HTTP serving removal.

Erethon:
> Deployed and seems to be working. Thank you for sending this patch!
> 
> ```
> $ curl -I http://cache.nixos.org/
> HTTP/1.1 301 Moved Permanently
> Connection: close
> Content-Length: 0
> Server: Varnish
> Retry-After: 0
> Location: https://cache.nixos.org/
> Accept-Ranges: bytes
> Date: Tue, 16 Apr 2024 19:23:05 GMT
> Via: 1.1 varnish
> X-Served-By: cache-vie6342-VIE
> X-Cache: HIT
> X-Cache-Hits: 0
> X-Timer: S1713295386.891547,VS0,VE0
> access-control-allow-origin: *
> ```

Can confirm it works, thank you for the quick review!

PR #418, opened 2024-04-23
go-neb: Only show Grafana link when applicable

Erethon:
This patch is a small change in the go-neb configuration to remove some clutter from the [infra alerts room](https://matrix.to/#/#infra-alerts:nixos.org). Since some alerts don't include a link to a Grafana dashboard, we can only print that part of the message when it's relevant.

I haven't tested this with an actual go-neb instance + alertmanager, but only a Go templating tester and seems to work fine.


delroth (approved):
Thank you! LGTM

PR #420, opened 2024-05-06
delft/hydra: remove authorized_keys for hydra users

mweinelt:
Nobody should ever need to directly log into these. The infra team already has root access and can switch to these users as needed.

### DIFF
--- a/delft/hydra.nix
+++ b/delft/hydra.nix
@@ -7,13 +7,6 @@ let
 in
 
 {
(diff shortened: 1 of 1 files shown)


vcunat (approved):
I'm puzzled why these lines were in there.  It's a very old piece, I didn't track beyond 15f4fd85.  Anyway, sounds good and unlikely to break stuff.

PR #425, opened 2024-05-15
rhea: more zramSwap, again

vcunat:
Three months ago the zram size config went from 50% to 80%, but now it doesn't seem enough again.


edef1c (approved):

mweinelt (approved):

mweinelt:
Applied.

PR #427, opened 2024-05-18
ssh-keys: cleanup

zimbatm:
* Remove unused keys
* Add missing members

### DIFF
--- a/ssh-keys.nix
+++ b/ssh-keys.nix
@@ -1,25 +1,16 @@
 rec {
-
   eelco = "ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAIAnI5L6oCgFyvEesL04LnbnH1TBhegq1Yery6TNlIRAA edolstra@gmail.com";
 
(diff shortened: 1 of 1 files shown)


ssh-keys.nix
```diff
@@ -1,48 +1,39 @@
...
-    delroth
     hexa-gaia
     hexa-helix
     vcunat
     zimbatm
   ];
 
   infra = infra-core ++ [
+    _999eagle-laptop
+    _999eagle-tower
```
vcunat, line 34: These two are missing definition, right?

JulienMalka (dismissed):
Adding people to `infra` is giving root access to `non-critical-infra` machines. Ideally, I'd not do that until people have been on-boarded to non-critical-infra and/or have started contributing to it. That being said, that's something I'd really want to see happen.

ssh-keys.nix
JulienMalka: I do see a definition, after `zimbatm`

vcunat: Now I also see and find it.  I'm not sure what I did wrong when I searched last time.

zimbatm:
Let me ask them.

zimbatm:
Let me make a separate PR for the additions

mweinelt (approved):

PR #429, opened 2024-05-22
ssh-keys: give linus access

zimbatm:
Finish the on-boarding so he can contribute to the project.


vcunat:
Sounds good to me.

mweinelt:
I feel like people should make the conscious effort to self-onboard. It is not a big ask.

PR #431, opened 2024-05-24
macs: add apple-virt, nixos-test system features

cole-h:


macs/nix-darwin.nix
```diff
@@ -37,6 +37,12 @@ in
     "extra-experimental-features" = [ "nix-command" "flakes" ];
     max-jobs = 4;
     cores = 2;
+    "system-features" = [
+      "apple-virt"
+      "nixos-test"
+      "big-parallel"
+      "nixos-test"
```
vcunat: One `"nixos-test"` should suffice 😄

cole-h: I don't know what you're talking about, there's definitely only one.... (nice catch, thanks, derp)

mweinelt:
I think this was tested, and you discussed it in the infra room, but I missed the outcome. Can someone bring me up to speed?

cole-h:
It was applied to one machine (`cosmic-stud`) and it "worked" in that they weren't being cancelled / aborted / whatever'd anymore, but the tests taking advantage of this would just hang (c.f. https://github.com/NixOS/nixpkgs/pull/303597#issuecomment-2130121108 and https://github.com/NixOS/nixpkgs/pull/303597#issuecomment-2135345269)

vcunat:
As it was always hanging, I rolled back the rhea's `machines` file, so that scheduler wouldn't try to run it.

Mic92:
Is this good to merge?

vcunat:
Running those jobs doesn't seem to work.  They hang and time out.

cole-h:
Closing. Feel free to reopen if we want to spend time figuring out why they're hanging.

PR #437, opened 2024-06-01
vaultwarden: use local mailer

### LINKED ISSUE #430: vaultwarden: smtp credentials invalid
**Affected service**
vault.nixos.org

**Describe the issue**
When inviting a user an error appears, that says SMTP authentication failed. An invitation is therefore never sent.

**System information**
 [2024-05-23 07:28:30.393][vaultwarden::mail][ERROR] SMTP 5xx error: permanent error (535): 5.7.8 Error: authentication failed:  - Authentication credentials invalid

JulienMalka:
Tried a test email with the admin interface
Fixes #430

### DIFF
--- a/non-critical-infra/modules/postfix.nix
+++ b/non-critical-infra/modules/postfix.nix
@@ -1,0 +1,48 @@
+{ config, pkgs, ... }:
+
+{
+  sops.secrets.opendkim-private-key = {
+    sopsFile = ../secrets/opendkim-private-key.caliban;
+    format = "binary";
+    owner = config.services.postfix.user;
+  };
+  services.opendkim = {
+    enable = true;
+    domains = config.networking.fqdn;
+    selector = "mail";
+    user = config.services.postfix.user;
+    group = config.services.postfix.group;
+    keyPath = "/run/opendkim-keys";
+  };
+
+  systemd.services.opendkim.serviceConfig = {
+    ExecStartPre = [
+      (
+        "+${pkgs.writeShellScript "opendkim-keys" ''
+          install -o ${config.services.postfix.user} -g ${config.services.postfix.group} -D -m0700 ${config.sops.secrets.opendkim-private-key.path} /run/opendkim-keys/${config.services.opendkim.selector}.private
+        ''}"
+      )
+    ];
+  };
+
+  services.postfix = {
+    enable = true;
+    hostname = config.networking.fqdn;
+    domain = config.networking.fqdn;
+    config = {
+      smtp_tls_note_starttls_offer = "yes";
+      smtp_tls_security_level = "may";
+      tls_medium_cipherlist = "AES128+EECDH:AES128+EDH";
+      smtpd_relay_restrictions = "permit_mynetworks permit_sasl_authenticated defer_unauth_destination";
+      mydestination = "localhost.$mydomain, localhost, $myhostname";
+      myorigin = "$mydomain";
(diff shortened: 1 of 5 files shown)


zimbatm (approved):
thanks!

zimbatm:
We're also going to have to set the SPF and DMARC records to avoid getting blacklisted.

```
vault.nixos.org.		300	IN	TXT	"v=spf1 a -all"
_dmarc.vault.nixos.org.	3483	IN	TXT	"v=DMARC1; p=none"
```
And also the record generated by  openkim-gen

JulienMalka:
> We're also going to have to set the SPF and DMARC records to avoid getting blacklisted.

ah yes will send tomorrow

non-critical-infra/modules/vaultwarden.nix
```diff
@@ -92,6 +92,52 @@
...
+          install -o ${config.services.postfix.user} -g ${config.services.postfix.group} -D -m0700 ${config.sops.secrets.opendkim-private-key.path} /run/opendkim-keys/${config.services.opendkim.selector}.private
+        ''}"
+      )
+    ];
+  };
+
+  services.postfix = {
+    enable = true;
+    hostname = "vault.nixos.org";
+    domain = "vault.nixos.org";
```
mweinelt, line 125:
```suggestion
    hostname = config.networking.fqdn;
    domain = config.networking.fqdn;
```

Let's generalize this into a `postfix.nix` and use the fqdn as the sender domain. Because it is very likely that we'll have other services on the same host wanting to send mail, and then service@fqdn would be the sender address of choice.

mweinelt:
DNS deployed.

non-critical-infra/modules/vaultwarden.nix
```diff
@@ -15,9 +16,9 @@
       ROCKET_ADDRESS = "127.0.0.1";
       ROCKET_PORT = 8222;
       ROCKET_LOG = "critical";
-      SMTP_HOST = "mail.gandi.net";
-      SMTP_PORT = 587;
-      SMTP_SSL = true;
+      SMTP_HOST = "localhost";
+      SMTP_PORT = 25;
+      SMTP_SSL = false;
       SMTP_FROM = "vaultwarden@nixos.org";
```
mweinelt, line 22:
```suggestion
      SMTP_FROM = "vaultwarden@caliban.nixos.org";
```

non-critical-infra/modules/postfix.nix
```diff
@@ -0,0 +1,48 @@
...
+    ];
+  };
+
+  services.postfix = {
+    enable = true;
+    hostname = config.networking.fqdn;
+    domain = config.networking.fqdn;
+    config = {
+      smtp_tls_note_starttls_offer = "yes";
+      smtp_tls_security_level = "dane";
```
mweinelt, line 34:
```
postfix/smtp[213546]: warning: DNSSEC validation may be unavailable
postfix/smtp[213546]: warning: reason: dnssec_probe 'ns:.' received a response that is not DNSSEC validated
```

We'd probably need to enable DNSSEC validation in resolved or set up a dedicated recursor. Not worth the hassle IMO.

mweinelt: I think the proper downgrade here would be `"may"`.

https://www.postfix.org/postconf.5.html#smtp_tls_security_level

PR #440, opened 2024-06-10
non-critical-infra/owncast: init

mweinelt:
Deploys owncast under live.nixos.org.


JulienMalka:
(Can't merge this bc needs a review from infra-build)

vcunat (approved):

vcunat:
I think that's just because of the DNS change.

vcunat:
I see DNS live already, nothing wrong here at a glance (and non-critical anyway).  So why block it?

PR #441, opened 2024-06-14
haumea/zrepl: reduce number of snapshots

mweinelt:
- 3x5m is not actionable since we often take more than 15 minutes to actually notice any issue
- 20x5m is a lot of write churn on the backup target, which is currently only my personal NAS
- 52x1w already offers 52 weekly backups, so 120x3w goes back 6 years and overlaps 17 times with 52x1w. Hopefully we don't need to restore data that is older than 1y.


vcunat (approved):
Yes.  I think this is good.  And fairly safe/conservative.

The `"120x3w"` part might help significantly with the size on your end (or other future receivers).

PR #447, opened 2024-06-23
haumea/zrepl: reduce snapshot count

vcunat:
Reduce snapshot count.  We repeatedly run out of space on Haumea.

### DIFF
--- a/build/haumea/zrepl.nix
+++ b/build/haumea/zrepl.nix
@@ -82,18 +82,26 @@
       };
 
       jobs = [
-        # XXX: Broken since 2024-01-10?
-        # (defaultBackupJob // {
-        #   name = "rsyncnet";
-        #   connect = {
-        #     identity_file = "/root/.ssh/id_ed25519";
-        #     type = "ssh+stdinserver";
-        #     host = "zh2543b.rsync.net";
-        #     user = "root";
-        #     port = 22;
-        #   };
-        # })
+        # Covers 20240629+
+        (defaultBackupJob // {
+          name = "rsyncnet";
+          connect = {
+            identity_file = "/root/.ssh/id_ed25519";
+            type = "ssh+stdinserver";
+            host = "zh4461b.rsync.net";
+            user = "root";
+            port = 22;
+          };
+        })
+        /* rsync.net provides a VM with FreeBSD
+           - almost nothing is preserved on upgrades except this "data1" zpool
+            $ scp ./zrepl.yml root@zh4461b.rsync.net:/usr/local/etc/zrepl/zrepl.yml
+            # pkg install zrepl
+            # service zrepl enable
+            # service zrepl start
+        */
 
+        /* Covered 2024: 0212 -- 0629
         (defaultBackupJob // {
           name = "hexa";
           connect = {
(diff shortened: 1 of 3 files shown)


build/haumea/zrepl.nix
```diff
@@ -33,23 +33,27 @@
       };
       pruning = {
         keep_sender = [
+          { type = "not_replicated"; }
```
vcunat, line 36: Oops, I expect we need have the `regex =` part here as well.

vcunat:
Let me dump a bit about why.

The problematic situation that we see have lots of data _unique_ to (some) in-between snapshots, i.e. dropping some of those snapshots (manually) could release lots of space.  Consequently:
 - making/transferring snapshots less often should decrease the total transfer amount and reduce this pressure on the remote
 - larger spacing between snapshots kept on Haumea should decrease the total space needed there, perhaps even if we didn't significantly decrease the total time span covered by snapshots

build/haumea/zrepl.nix
vcunat: Or maybe not.  Their examples set doesn't have it:
https://zrepl.github.io/configuration/prune.html#pruning-policies

vcunat:
🤔  as for the backup location(s), it feels wasteful to keep every week uniformly for a year.  Can you see any reason for it?  I'd intuitively again go for some exponentially increasing spacing.  I assume we can afford more space than on Haumea itself, so e.g. this slower one?
```nix
              "2x1h"
              "2x2h"
              "2x4h"
              "4x8h"
              # At this point the grid spans 2 days (-2h) by 10 snapshots.
              # (See note above about 8h -> 24h.)
              "2x1d"
              "2x2d"
              "2x4d"
              "2x8d"
              "2x16d"
              "2x32d"
              "2x64d"
              "2x128d"
              # At this point we keep 26 snapshots spanning 384--512 days (depends on moment),
              # with exponentially increasing spacing (almost).
```

Perhaps note the [docs](https://zrepl.github.io/configuration/prune.html#policy-grid) that the specified intervals do not overlap.  All the intervals are stacked in the specified order and multiplicity, forming a fixed grid.

build/haumea/zrepl.nix
```diff
@@ -33,23 +33,34 @@
...
             grid = lib.concatStringsSep " | " [
-              "4x15m"
-              "24x1h"
-              "4x1d"
-              "3x1w"
+              "1x1h"
+              "1x2h"
+              "1x4h"
+              # "grid" acts weird if an interval isn't a whole-number multiple
+              # of the previous one, so we jump from 8h to 24h
```
vcunat, line 45: Not sure if it's worth trying to explain the weirdness.  I base it not on actual experience but on definition in their docs &ndash; and how such a model behaves then when running continuously.

build/haumea/zrepl.nix
vcunat: Maybe we should drop this line anyway.  In case the remote end is down, we probably want to keep pruning the sender to reduce the risk of running out of space.

zimbatm (approved):
LGTM!

PR #453, opened 2024-07-18
hydra: increase local worker thread count

mweinelt:
Our hydra fork has a limit on the number of active build slots, which defaults to the number of threads minus two, so 32-2=30.

The plan is to increase the utilization of rhea more to unblock build slots on the various builders.


zimbatm (approved):

mweinelt:
As rhea is throttling at the current thread count this does not make sense after all.

PR #456, opened 2024-08-17
disko cleanup + deadnix

Mic92:
It's disko's job to add filesystems, so no need to duplicate this

After the change:

nix-repl> nixosConfigurations."caliban.nixos.org".config.boot.supportedFilesystems { zfs = true; }


zimbatm (approved):
LGTM

PR #460, opened 2024-08-18
terraform: remove plugins without the opentofu prefix

Mic92:


Mic92:
Also part of https://github.com/NixOS/infra/pull/462

PR #474, opened 2024-09-12
ssh-keys: remove inactive users

Mic92:
those keys are not used but also their owner no longer seem active in nixos-infra.


vcunat (approved):
OK, why not.  (I haven't really checked that the keys aren't referenced anywhere.)

PR #492, opened 2024-10-15
Introduce staging binary cache to reduce our binary cache size

Mic92:
The idea is to have buckets to test VCL configuration on them.
In particular I want to create a new S3 bucket that we would start pushing to with hydra.
The VCL rules in fastly would check both buckets for the migration period.
After we have migrated the part of the binary cache that we are interested,
we make the VCL rules only hit the new bucket.
After that we add an archiving rule to transition the old bucket to aws glacier: https://aws.amazon.com/blogs/aws/archive-s3-to-glacier/


edolstra:
What's this migration all about?

Mic92:
@edolstra to reduce AWS costs on our binary cache.

edolstra:
@Mic92 Yeah but why do we need a separate bucket for that? The way I imagine doing GC is that the GC process would set a tag on objects that we consider garbage, and then we can have a lifecycle rule that moves them to Glacier. As an intermediate step, we could have an access control rule that only makes them inaccessible (so if we screw up, we can restore them more cheaply than from Glacier).

Mic92:
@edolstra sounds easier actually. I thought it might be also expensive to run api requests on all objects. But it looks like it's still doable:

Let's say we need 1-2 requests per object it would be ({1,2} * 889587727) / (1e6) = 889.5 -> 1779.1
What would be the best way to find all the objects to keep.

Mic92:
Okay, so I also had a chat with @edef1c who is working on a more sound gc algorithm.
@zimbatm also documented the current tools and what needs to be done (hopefully soon documented in a more public state)
From that I saw that it's actually not so trivial to get the list of all live store paths for this big bucket.
So here is my plan B in case we cannot make the gc work soon enough:

1. We would just deploy this solution and make hydra push to the new bucket. Old store paths would be accessible through the old bucket and new data in the new bucket. Fastly will check both.
2. We remove the old bucket from fastly after some transition period.
3. After some testing period, we send the whole old bucket to glacier.

If someone wants to suggest a technically better solution, they are welcome to implement it.

Mic92:
I think this staging infrastructure is already useful on its own to test new configuration with fastly without breaking cache.nixos.org. So I am going to merge this.

edolstra:
> Let's say we need 1-2 requests per object it would be ({1,2} * 889587727) / (1e6) = 889.5 -> 1779.1 Dollar

That's pretty acceptable as a one-time cost.

>  What would be the best way to find all the objects to keep?

I once wrote [a `nix` command](https://github.com/edolstra/nix/commit/a1f84a33f79a9ea9ad18c99608ccdd9052e43c1f) that determines the closure of all reachable store paths from the `store-paths.xz` files of the NixOS/Nixpkgs releases (and whatever other roots we might want to keep).

Mic92:
@edolstra great. We can use consider this command if someone also provides the actual GC scripts. Does this run the check single threaded? How long do you think it would take to process a single eval? @edef1c also pointed out that tagging s3 objects can be potential dangerous because hydra might start referring to the objects that we considered dead. She was proposing to enable versioning and delete the current version of the narinfo instead so that hydra would re-create it. I will proceed testing my two-bucket option further with a staging hydra, but please don't feel blocked on improving your gc alternative. We should certainly use it if it's ready in time.

Issue #485, opened 2024-09-30
Migrate away from ImprovMX for mailing lists


jfly:
We currently use [ImprovMX](https://improvmx.com/) to handle mail sent to `@nixos.org` (see relevant [dns entries](https://github.com/nixos/infra/blob/c302b98f418757802f26a3cab9ce252ecdf1d305/terraform/dns.tf#L358-L370)).

- We only use ImprovMX for mail forwarding (teams like `infra@`, `marketing@`, etc). Today, nobody sends mail from `@nixos.org`, and nobody has any inboxes.
- You need a web account with ImprovMX to see and to update these mail forwards. The Nix community can't see/audit any of this.
- There are various limits (number of forwards, perhaps the number of emails an address can forward to?). See https://improvmx.com/pricing/. I don't know if we're currently paying for ImprovMX. I think I heard that we've run into some of these limits.

## The plan

A few weeks ago, [@Mic92 asked me to look into self hosting this instead](https://pad.lassul.us/nixos-infra?view#2024-09-12). He recommended [Simple NixOS Mailserver](https://gitlab.com/simple-nixos-mailserver/nixos-mailserver) (SNM). I've played with it a bit, and it does seem like a good fit here.

1. [x] Install SNM on [umbriel](https://github.com/NixOS/infra/tree/master/non-critical-infra/hosts/umbriel.nixos.org).
   - The configuration docs here are great: https://nixos-mailserver.readthedocs.io/en/latest/setup-guide.html.
   - Leave `mailserver.loginAccounts` empty, and disable [pop](https://nixos-mailserver.readthedocs.io/en/latest/options.html#cmdoption-arg-mailserver.enablePop3)/[imap](https://nixos-mailserver.readthedocs.io/en/latest/options.html#cmdoption-arg-mailserver.enableImap).
   - Port the existing mailing lists from ImprovMX to [`mailserver.forwards`](https://nixos-mailserver.readthedocs.io/en/latest/options.html#cmdoption-arg-mailserver.forwards)
     - @Mic92 has posted a dump [REMOVED] (accurate as of 2024-09-30).
2. [x] Verify this server can successfully send mail (target: 10/10 on https://www.mail-tester.com/). Either by temporarily adding a login account, or speaking directly to postfix via the cli.
4. [x] Monitor smtp tls (see below).
5. [x] Alert on stmp tls monitor failures.
[...]

PR #495, opened 2024-10-29
Add simple-nixos-mailserver to umbriel

jfly:
This is part 1 of https://github.com/NixOS/infra/issues/485

### DIFF
--- a/non-critical-infra/flake-module.nix
+++ b/non-critical-infra/flake-module.nix
@@ -26,6 +26,7 @@
             value
             inputs.disko.nixosModules.disko
             inputs.first-time-contribution-tagger.nixosModule
+            inputs.simple-nixos-mailserver.nixosModule
             inputs.sops-nix.nixosModules.sops
           ];
           extraModules = [ inputs.colmena.nixosModules.deploymentOptions ];
@@ -49,8 +50,14 @@
     };
 
   perSystem =
-    { pkgs, inputs', ... }:
+    { inputs', ... }:
+    # Use the latest packages from `nixpkgs-unstable` for dev tools.
+    let
+      pkgs = inputs'.nixpkgs-unstable.legacyPackages;
+    in
     {
+      packages.encrypt-email-address = pkgs.callPackage ./packages/encrypt-email-address { };
+
       devShells.non-critical-infra = pkgs.mkShellNoCC {
         packages = [
           inputs'.colmena.packages.colmena
--- a/non-critical-infra/modules/mailserver/default.nix
+++ b/non-critical-infra/modules/mailserver/default.nix
@@ -1,0 +1,19 @@
+{ config, ... }:
+
+{
+  imports = [ ./mailing-lists.nix ];
+
+  mailserver = {
+    enable = true;
+    certificateScheme = "acme-nginx";
+
+    # Until we have login accounts, there's no reason to run either of these.
+    enablePop3 = false;
+    enableImap = false;
+
+    fqdn = config.networking.fqdn;
+
+    # TODO: change to `nixos.org` when ready
+    domains = [ "mail-test.nixos.org" ];
+  };
+}
--- a/non-critical-infra/modules/mailserver/mailing-lists.nix
+++ b/non-critical-infra/modules/mailserver/mailing-lists.nix
@@ -1,0 +1,64 @@
+# This module provides the mailing list definitions for `@nixos.org`.
+#
+# Simply change the `lists` attribute set below to create new mailing lists or
+# edit membership of existing lists.
+#
+# If you wish to hide your email address, you can encrypt it with SOPS. Just
+# run `nix run .#encrypt-email-address -- --help` and follow the instructions.
+
+{ config, lib, ... }:
+
+let
+  # Mailing lists go here.
+  # TODO: replace with the real `nixos.org` mailing lists.
+  listsWithSecretFiles = {
+    "test-list@mail-test.nixos.org" = [
+      "jfly@playground.jflei.com"
+      ../../secrets/jfly-email.umbriel
+      "jeremyfleischman+subscriber@gmail.com"
+    ];
+  };
+
+  fileToSecretId = file: builtins.baseNameOf file;
+
+  listsWithSecretPlaceholders = lib.mapAttrs' (name: members: {
+    name = name;
+    value = map (
+      member:
+      if builtins.isString member then member else config.sops.placeholder.${fileToSecretId member}
+    ) members;
+  }) listsWithSecretFiles;
+
+  secretFiles = lib.pipe listsWithSecretFiles [
+    (lib.mapAttrsToList (_name: members: members))
+    lib.flatten
+    (builtins.filter (member: !builtins.isString member))
+  ];
+in
+
+{
+  # Declare secrets for every secret email in the lists above.
+  sops.secrets = builtins.listToAttrs (
+    map (file: {
+      name = fileToSecretId file;
+      value = {
+        format = "binary";
+        sopsFile = file;
+      };
+    }) secretFiles
+  );
+
+  # Whenever this changes, we need to manually restart the `postfix-setup`
+  # service for postfix to notice the change.
+  # TODO: <https://github.com/NixOS/infra/issues/505> tracks fixing this
+  sops.templates."postfix-virtual-mailing-lists".content = lib.concatStringsSep "\n" (
+    lib.mapAttrsToList (
+      name: members: "${name} ${lib.concatStringsSep ", " members}"
+    ) listsWithSecretPlaceholders
+  );
+
+  services.postfix.mapFiles.virtual-mailing-lists =
+    config.sops.templates."postfix-virtual-mailing-lists".path;
+
+  services.postfix.config.virtual_alias_maps = [ "hash:/etc/postfix/virtual-mailing-lists" ];
+}
--- a/non-critical-infra/packages/encrypt-email-address/default.nix
+++ b/non-critical-infra/packages/encrypt-email-address/default.nix
@@ -1,0 +1,20 @@
+{
+  lib,
+  python3,
+  sops,
+}:
+
+python3.pkgs.buildPythonApplication {
+  name = "encrypt-email-address";
+  src = ./.;
+
+  format = "other";
+
+  propagatedBuildInputs = [ python3.pkgs.click ];
+
+  installPhase = ''
+    mkdir -p $out/bin
+    mv ./encrypt-email-address.py $out/bin/encrypt-email-address
+    wrapProgram $out/bin/encrypt-email-address --prefix PATH : ${lib.makeBinPath [ sops ]}
+  '';
+}
--- a/non-critical-infra/packages/encrypt-email-address/encrypt-email-address.py
+++ b/non-critical-infra/packages/encrypt-email-address/encrypt-email-address.py
@@ -1,0 +1,101 @@
+#!/usr/bin/env python3
+
+import re
+import subprocess
+from pathlib import Path
+
+import click
+
+
+def find_project_root(start: Path) -> Path:
+    # Can search for `flake.nix` because there are multiple in this project.
+    root_indicator = start / ".git/config"
+    if root_indicator.exists():
+        return start
+
+    return find_project_root(start.parent)
+
+
+@click.command()
+@click.argument("address_id")
+@click.argument("email")
+@click.option("--force/--no-force", "-f/ ", default=False)
+def main(address_id: str, email: str, force: bool) -> None:
+    """
+    Encrypt an email address (or email addresses) for inclusion in a mailing list.
+
+    Example:
+
+        \bencrypt-email-address some-token 'me@example.com,you@example.com'
+
+    Then follow the instructions for what to do next.
+    """
+    # Feel free to make the regex less restrictive if you need to.
+    id_re = re.compile("[A-Za-z0-9-]+")
+    if not id_re.fullmatch(address_id):
+        msg = f"Given ID: {address_id!r} is invalid. Must match regex: {id_re.pattern}"
+        raise click.ClickException(msg)
+
+    # Make sure we aren't being given a text file that happens to have a newline at the end.
+    clean_email = email.strip()
+    if clean_email != email:
+        click.secho("Removed whitespace surrounding given email address", fg="yellow")
+    email = clean_email
+
+    project_root = find_project_root(Path.cwd()).relative_to(Path.cwd(), walk_up=True)
+    non_critical_infra_dir = project_root / "non-critical-infra"
+
+    secret_path = non_critical_infra_dir / f"secrets/{address_id}-email.umbriel"
+
+    if secret_path.exists():
+        if not force:
+            msg = f"Refusing to clobber existing {secret_path}. Use `--force` to override."
+            raise click.ClickException(msg)
+        click.secho(f"Clobbering existing {secret_path}", fg="yellow")
+
+    sops_config = non_critical_infra_dir / ".sops.yaml"
+    cp = subprocess.run(
+        [
+            "sops",
+            "--encrypt",
+            "--config",
+            sops_config,
+            "--filename-override",
+            secret_path,
+            "/dev/stdin",
+        ],
+        text=True,
+        check=True,
+        stdout=subprocess.PIPE,
+        input=email,
+    )
+
+    secret_path.write_text(cp.stdout)
+    subprocess.run(
+        ["git", "add", "--intent-to-add", "--force", "--", secret_path], check=True
+    )
+
+    click.secho(f"Successfully generated {secret_path}", fg="green")
+
+    mailing_list_nix = non_critical_infra_dir / "modules/mailserver/mailing-lists.nix"
+    assert mailing_list_nix.exists()
+
+    click.secho()
+    click.secho("Now add yourself to ", nl=False)
+    click.secho(mailing_list_nix, fg="blue", nl=False)
+    click.secho(". ")
+
+    click.secho()
+    click.secho("Lastly, add `", nl=False)
+    click.secho(
+        secret_path.relative_to(mailing_list_nix.parent, walk_up=True),
+        fg="blue",
+        nl=False,
+    )
+    click.secho("` to the relevant mailing list under '", nl=False)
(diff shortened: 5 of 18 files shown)


non-critical-infra/hosts/umbriel.nixos.org/mailserver/default.nix
```diff
@@ -0,0 +1,34 @@
...
+    # Keep in sync with `iac/pulumi/app/dns.py`
+    fqdn = config.networking.fqdn;
+    # TODO: create an `MX` record for `playground.nixos.org` (value `umbriel.nixos.org`): <https://nixos-mailserver.readthedocs.io/en/latest/setup-guide.html#set-a-mx-record>.
+    # TODO: create an `SPF` record for `playground.nixos.org`: <https://nixos-mailserver.readthedocs.io/en/latest/setup-guide.html#set-a-mx-record>.
+    # TODO: create `DKIM` TXT record: <https://nixos-mailserver.readthedocs.io/en/latest/setup-guide.html#set-dkim-signature>.
+    #       (can't do this until after SNM is deployed)
+    # TODO: set a `DMARC` record: <https://nixos-mailserver.readthedocs.io/en/latest/setup-guide.html#set-a-dmarc-record>.
+    #       (and then make it strict (`v=DMARC1; p=reject; adkim=s; aspf=s;`) once things are working)
+    # TODO: change to `nixos.org` when ready
+    domains = [ "playground.nixos.org" ];
```
jfly, line 21: @Mic92, I was thinking we could first set this up end to end with some other domain than `nixos.org` so we can test it out without breaking existing `nixos.org` mail. Perhaps you have a different idea for testing this out?

non-critical-infra/hosts/umbriel.nixos.org/mailserver/default.nix
```diff
@@ -0,0 +1,34 @@
...
+
+  services.postfix.config.virtual_alias_maps = [
+    # Expanding https://github.com/NixOS/nixpkgs/blob/a9fe4d6d8ccde780e872ed1446f3746498152663/nixos/modules/services/mail/postfix.nix#L833C87-L833C108
+    "hash:/etc/postfix/virtual-mailing-lists"
+  ];
+
+  # TODO: replace with the real `nixos.org` mailing lists.
+  # TODO: port this to be a [sops-nix template](https://github.com/Mic92/sops-nix?tab=readme-ov-file#templates)
+  services.postfix.mapFiles.virtual-mailing-lists = pkgs.writeText "postfix-virtual-mailing-lists" ''
+    jfly@playground.nixos.org jfly@playground.jflei.com, jeremyfleischman+subscriber@gmail.com
```
jfly, line 32: Again, just an idea for testing this out.

Mic92:
I think we only need to provide SMTP logins in special circumstances i.e. for the people that do the vote. Otherwise people can respond with their own email address for transparency reasons and use the alias in cc in case there are multiple people.

Mic92:
For the sops integration, we should have an ad-hoc sops script that creates one file per recipient. Reason is that the contributor won't have access to anything but the public keys of the admin team and our server. This way they can encrypt but not decrypt.

jfly:
> we should have an ad-hoc sops script that creates one file per recipient

I've added a new `nix run .#encrypt-email-address` command for this. People will discover this command if they read the documentation I added in `non-critical-infra/hosts/umbriel.nixos.org/mailserver/README.md`

non-critical-infra/packages/encrypt-email-address/encrypt-email-address.py
```diff
@@ -0,0 +1,100 @@
...
+        Encrypt an email address (or email addresses) for inclusion in a mailing list.
+
+        See `non-critical-infra/modules/mailserver/mailing-lists.nix` for
+        documentation about how to use this.
+    """)
+)
+@click.argument("id")
+@click.argument("email")
+@click.option("--force/--no-force", "-f/ ", default=False)
+def main(id: str, email: str, force: bool):
```
Mic92, line 22:
```suggestion
def main(id: str, email: str, force: bool) -> None:
```

non-critical-infra/modules/mailserver/mailing-lists.nix
```diff
@@ -0,0 +1,39 @@
...
+      "jeremyfleischman+subscriber@gmail.com"
+    ];
+  };
+in
+
+{
+  # Encrypted email addresses go here.
+  sops.secrets.jfly-email = {
+    format = "binary";
+    sopsFile = ../../secrets/jfly-email.umbriel;
```
Mic92, line 28: Maybe we want to put this in some subdirectory and import all files in it using `builtins.readDir`. Than adding new emails will be easier.

non-critical-infra/packages/encrypt-email-address/encrypt-email-address.py
```diff
@@ -0,0 +1,100 @@
...
+
+    cp = subprocess.run(
+        ["sops", "--encrypt", "--filename-override", secret_path, "/dev/stdin"],
+        text=True,
+        check=True,
+        stdout=subprocess.PIPE,
+        input=email,
+    )
+
+    secret_path.write_text(cp.stdout)
```
Mic92, line 59: This command would be handy:

```
  git add --intent-to-add --force -- {secret_path}
```

non-critical-infra/packages/encrypt-email-address/default.nix
```diff
@@ -0,0 +1,24 @@
...
+  src = ./.;
+
+  format = "other";
+
+  nativeBuildInputs = [ ruff ];
+
+  propagatedBuildInputs = [ python3.pkgs.click ];
+
+  installPhase = ''
+    ruff check ./encrypt-email-address.py
```
SuperSandro2000, line 19: Do we really want this to potentially fail to build when ruff is being updated?

non-critical-infra/packages/encrypt-email-address/encrypt-email-address.py
```diff
@@ -0,0 +1,100 @@
+#!/usr/bin/env python
```
SuperSandro2000, line 1:
```suggestion
#!/usr/bin/env python3
```

non-critical-infra/packages/encrypt-email-address/encrypt-email-address.py
```diff
@@ -0,0 +1,100 @@
...
+    if not id_re.fullmatch(id):
+        raise click.ClickException(
+            f"Given ID: {id!r} is invalid. Must match regex: {id_re.pattern}"
+        )
+
+    # Make sure we aren't being given a text file that happens to have a newline at the end.
+    if "\n" in email:
+        raise click.ClickException(
+            "Email address must not not contain a newline character"
+        )
```
SuperSandro2000, line 34: This is kinda silly. Most editors append a final new line in linux land. Can't we rather strip whitespace?

non-critical-infra/packages/encrypt-email-address/encrypt-email-address.py
```diff
@@ -0,0 +1,100 @@
...
+
+    if Path.cwd().name != "non-critical-infra":
+        raise click.ClickException(
+            "You must run this command inside the `non-critical-infra/` folder."
+        )
```
SuperSandro2000, line 39: Just changedir to the correct directory?

non-critical-infra/packages/encrypt-email-address/default.nix
Mic92: Yes, we control the nixpkgs instance and the source here, so we should address new errors. However we already do this check in https://github.com/NixOS/infra/blob/master/formatter/flake-module.nix so this seem a bit redundant.

```suggestion
```

non-critical-infra/packages/encrypt-email-address/encrypt-email-address.py
Mic92: Isn't this email passed by cli rather than editor? But I would be also interested how we end up with a newline here.

non-critical-infra/modules/mailserver/mailing-lists.nix
```diff
@@ -0,0 +1,39 @@
...
+# run `nix run .#encrypt-email-address` in the `non-critical-infra/` folder and
+# follow the instructions.
+
+{ config, lib, ... }:
+
+let
+  # Mailing lists go here.
+  # TODO: replace with the real `nixos.org` mailing lists.
+  lists = {
+    "jfly@playground.nixos.org" = [
```
Mic92, line 16: If you haven't seen this pull request, you may not know what `playground` means. `test-mail` makes it very obvious:

```suggestion
    "jfly@test-mail.nixos.org" = [
```

non-critical-infra/modules/mailserver/default.nix
```diff
@@ -0,0 +1,25 @@
...
+    enableImap = false;
+
+    fqdn = config.networking.fqdn;
+
+    # TODO: create an `MX` record for `playground.nixos.org` (value `umbriel.nixos.org`): <https://nixos-mailserver.readthedocs.io/en/latest/setup-guide.html#set-a-mx-record>.
+    # TODO: create an `SPF` record for `playground.nixos.org`: <https://nixos-mailserver.readthedocs.io/en/latest/setup-guide.html#set-a-mx-record>.
+    # TODO: create `DKIM` TXT record: <https://nixos-mailserver.readthedocs.io/en/latest/setup-guide.html#set-dkim-signature>.
+    #       (can't do this until after SNM is deployed)
+    # TODO: set a `DMARC` record: <https://nixos-mailserver.readthedocs.io/en/latest/setup-guide.html#set-a-dmarc-record>.
+    #       (and then make it strict (`v=DMARC1; p=reject; adkim=s; aspf=s;`) once things are working)
```
Mic92, line 21:
```suggestion
    #       (and maybe make it strict (`v=DMARC1; p=reject; adkim=s; aspf=s;`) after checking if this is not breaking mailing lists)
```

non-critical-infra/modules/mailserver/mailing-lists.nix
jfly: I totally agree. I've reworked this. It's not quite what you suggest: instead now the list of mailing list members contains either strings or files. I then extract the files from the member lists to declare the sops secrets.

jfly (commented):
Thanks for the feedback!

non-critical-infra/packages/encrypt-email-address/default.nix
jfly: > However we already do this check in https://github.com/NixOS/infra/blob/master/formatter/flake-module.nix so this seem a bit redundant

That's `ruff-format`, not `ruff-check`. But good point! I've added that over in https://github.com/NixOS/infra/pull/498

jfly: I've removed the call to `ruff check` (in favor of #498). I'm marking this resolved.

non-critical-infra/packages/encrypt-email-address/encrypt-email-address.py
jfly: You're right @Mic92. The scenario I was vaguely worried about was somebody passing a file as a string, like this:

```
nix run .#encrypt-email-address jfly "$(< /tmp/foo)"
```

I'm also worried about the scenario where someone uses `sops edit` on these files directly, which will likely introduce a newline, and I can't think of anything clever we can do about it. Do y'all have any bright ideas?

non-critical-infra/modules/mailserver/mailing-lists.nix
jfly: Updated!

non-critical-infra/modules/mailserver/default.nix
jfly: Done!

non-critical-infra/flake-module.nix
```diff
@@ -49,8 +50,14 @@
     };
 
   perSystem =
-    { pkgs, inputs', ... }:
+    { inputs', ... }:
+    # Use the latest packages from `nixpkgs-unstable` for dev tools.
```
jfly, line 54: It's entirely non-obvious from this PR: I need sops `--filename-override` flag, which was introduced in [sops 3.9.0], which is not present in the stable nixos release we use here.

[sops 3.9.0]: https://github.com/getsops/sops/blob/main/CHANGELOG.rst#390

non-critical-infra/packages/encrypt-email-address/encrypt-email-address.py
Mic92: I think this is based on suggestions of this pep: https://peps.python.org/pep-0394/

non-critical-infra/packages/encrypt-email-address/encrypt-email-address.py
Mic92: You can calculate the script directory using `Path(__name__).parent` and than compute the path you want to patch relative to that.

non-critical-infra/packages/encrypt-email-address/encrypt-email-address.py
Mic92: Mhm. Can we not just do a strip() instead to remove all whitespace around the string?

non-critical-infra/packages/encrypt-email-address/encrypt-email-address.py
SuperSandro2000: Just to be sure no one ever tries to run the script with python2

non-critical-infra/packages/encrypt-email-address/encrypt-email-address.py
jfly: Done.

terraform/dns.tf
```diff
@@ -330,7 +330,28 @@ locals {
...
+    # Mailserver configuration for `mail-test.nixos.org`
+    {
+      hostname = "mail-test.nixos.org"
+      type     = "MX"
+      value    = "umbriel.nixos.org"
+    },
+    {
+      hostname = "wiki.nixos.org"
+      type     = "TXT"
+      value    = "v=spf1 a:mail-test.nixos.org -all"
```
jfly, line 344: @Mic92, I see we're hardcoding IPs here: https://github.com/NixOS/infra/blob/fb5e9aeb6c1621c98d2c9ba029613f3be7c9e1db/terraform/dns.tf#L317, but these docs use a domain name: https://nixos-mailserver.readthedocs.io/en/latest/setup-guide.html#set-a-spf-record.

Any reason to do the IPs rather than a domain name?

Mic92: `a:` seems to be ipv4 only? We also have ipv6.

jfly: Despite the name, I think it's dual ipv4/ipv6. From https://datatracker.ietf.org/doc/html/rfc7208#section-5.3:

> An address lookup is done on the <target-name> using the type of lookup (A or AAAA) appropriate for the connection type (IPv4 or IPv6)

[From experience](https://gitlab.com/simple-nixos-mailserver/nixos-mailserver/-/merge_requests/342), I know that my personal mailserver is (now) successfully sending emails to Gmail over IPv6.

jfly:
I'm deploying this to umbriel now

jfly:
@Mic92, this PR is updated with the DKIM TXT record. I've also added a README to umbriel explaining that this will need to be updated if we reprovision the machine from scratch.

jfly:
@Mic92, this PR is updated with the DKIM TXT record. I've also added a README to umbriel explaining that this will need to be updated if we reprovision the machine from scratch.

cole-h:
Just a small note: there should probably be some kind of monitoring before this sees more widespread use. Back when I ran my own instance of SNM, something (I don't remember what, it's been years at this point) went wrong and I never noticed that I stopped receiving mail (I didn't use this domain a lot at the time).

terraform/dns.tf
jfly: @mweinelt:

1. I like the thought of using `mx` rather than `a`.
2. I'm not sure about the `~all` vs `-all`. I'd be inclined to keep it as strict as possible until we have a good reason not to.

mweinelt: Yeah, I prefer strictness as well.

jfly: I've confirmed with my personal test mailserver that `mx` seems to work here. I've changed this SPF record to `v=spf1 mx -all`.

I'm going to mark this resolved.

non-critical-infra/packages/encrypt-email-address/encrypt-email-address.py
jfly: I've implemented 1) by searching upwards for a `.git/config` directory. It was pretty simple to do: https://github.com/NixOS/infra/pull/495/commits/240fff84baebf4372d72bd4967ce30bb27ae1638. Now you can run this command from anywhere in the project =)

I'm going to mark this resolved.

non-critical-infra/packages/encrypt-email-address/encrypt-email-address.py
```diff
@@ -0,0 +1,100 @@
...
+        Encrypt an email address (or email addresses) for inclusion in a mailing list.
+
+        See `non-critical-infra/modules/mailserver/mailing-lists.nix` for
+        documentation about how to use this.
+    """)
+)
+@click.argument("user")
+@click.argument("email")
+@click.option("--force/--no-force", "-f/ ", default=False)
+def main(user: str, email: str, force: bool) -> None:
```
jfly, line 31: @Mic92, I don't think "user" is really the right name for this. Perhaps `address_id`?

Mic92:
@jfly I applied your terraform changes. Let me know when this is good to merge.

non-critical-infra/packages/encrypt-email-address/encrypt-email-address.py
jfly: Done!

Mic92 (approved):

mweinelt:
Any reason this was just merged after #505 was opened?

jfly:
We wanted to get this merged because the ruff linter changes felt conflict-prone: <https://github.com/NixOS/infra/commit/8e5be8674f19dd0d38d25ef1cb64b24582e237c3>.

Mic92:
@mweinelt I didn't wanted the diff get too large. And given this service is just running on mail-test.nixos.org, the issue mentioned doesn't have any real-world impact on us.

PR #508, opened 2024-11-09
Customize bounce template for mailserver

jfly:
The default template tells you to "please send mail to postmaster" which isn't very actionable (what's the postmaster?). This directs people to a place where they can actually get help.

# Before

2024-11-09_11-48-13_pattern

# After

2024-11-09_12-25-36_pattern

# Alternatives considered

- We could set up a mailing list for the infra team, and direct people to it
- We could direct people to reach out to the infra team over matrix

### DIFF
--- a/non-critical-infra/modules/mailserver/default.nix
+++ b/non-critical-infra/modules/mailserver/default.nix
@@ -1,4 +1,4 @@
-{ config, ... }:
+{ config, pkgs, ... }:
 
 {
   imports = [ ./mailing-lists.nix ];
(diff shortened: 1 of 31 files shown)


non-critical-infra/modules/mailserver/default.nix
```diff
@@ -16,4 +16,50 @@
...
+    Subject: Undelivered Mail Returned to Sender
+    Postmaster-Subject: Postmaster Copy: Undelivered Mail
+
+    This is the mail system at host $myhostname.
+
+    I'm sorry to have to inform you that your message could not
+    be delivered to one or more recipients. It's attached below.
+
+    For further assistance, please file an issue at
+    https://github.com/NixOS/infra/issues/new
```
jfly, line 33: A downside with this is that it will likely result in people revealing the problematic email addresses publicly. @mweinelt, do you have a suggestion here?

mweinelt: I think GitHub issues is indeed the best contact point for load-balancing this. Whoever submits these needs to anonymize the relevant logs as needed. If they don't do it, we still can and we can even delete the edit history.

jfly: Sounds great. I've updated the templates to instruct people to anonymize personal email addresses in their report.

Mic92 (approved):

PR #509, opened 2024-11-09
Use a looser SPF record

jfly:
Unfortunately, this doesn't play nicely with mailing lists and some receiving mail architectures.

See
https://gitlab.com/simple-nixos-mailserver/nixos-mailserver/-/issues/302 for a complete explanation.

### DIFF
--- a/terraform/dns.tf
+++ b/terraform/dns.tf
@@ -341,7 +341,7 @@ locals {
     {
       hostname = "mail-test.nixos.org"
       type     = "TXT"
-      value    = "v=spf1 mx -all"
+      value    = "v=spf1 mx ~all"
     },
     {
       hostname = "mail._domainkey.mail-test.nixos.org"


mweinelt:
Isn't that really the problem of the list provider? They need to rewrite the headers or wrap the original message.

jfly:
My understanding is that list providers shouldn't do any "munging" of the message if DKIM is set up. The idea is that yes, SPF will fail, and that's fine, because DKIM will still pass, and you only need 1 to pass for DMARC to pass. See <https://begriffs.com/posts/2018-09-18-dmarc-mailing-list.html#if-dkim-is-used> for details.

zimbatm (approved):

SuperSandro2000:
> Isn't that really the problem of the list provider? They need to rewrite the headers or wrap the original message.

I was under the same impression. Especially for domains which have no DKIM/DMARC.

jfly:
For the record, @mweinelt and @SuperSandro2000 were both correct here: we need to rewrite the email envelope so mailservers we're forwarding to get a SPF success. This mechanism is called Sender Rewriting Scheme (SRS), and we implemented it in https://github.com/NixOS/infra/pull/622.

PR #515, opened 2024-11-29
channels: Promote NixOS 24.11 to stable

RossComputerGuy:
*To be released on Saturday, November 30th.*

### DIFF
--- a/channels.nix
+++ b/channels.nix
@@ -42,33 +42,33 @@ rec {
     "nixos-24.11" = {
       job = "nixos/release-24.11/tested";
       variant = "primary";
-      status = "beta";
+      status = "stable";
     };
     "nixos-24.11-small" = {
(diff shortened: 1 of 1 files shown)


fricklerhandwerk:
Maybe someone will be so kind to [run an update on the Nixpkgs releases for nix.dev](https://github.com/NixOS/nix.dev/blob/master/CONTRIBUTING.md#updating-reference-manuals) so the manuals show up correctly, in case I don't get to it. 

There's currently no automation because noone had the time to build it, but also because those updates are quite expensive, so it doesn't make much sense to run them periodically, and making it push-based requires more stars to align.

mweinelt:
That is outside of what infra manages.

RossComputerGuy:
This is ready to merge.

mweinelt (approved):

PR #519, opened 2024-12-10
build: prune common configuration

mweinelt:
- Remove hard dependency on coretemp
- Remove graphics workaround for old environment at TU Delft
- Update available system packages for better workability
- Group options logically and sort them from low-level to high-level
- Remove bespoke /tmp tmpfiles option, nixpkgs has its own with 10d cleanup
- Remove builder specific config (KSM, build users, login limits)
- Remove cron email config leftover, we don't use cron timers anymore

This is just the first commit. See individual commit messages for remaining changes.


Mic92:
@mweinelt since you seem to have done some terraform provider migration to opentofu, can you also add this to your pull request: https://github.com/NixOS/infra/pull/520 ?

build/common.nix
```diff
@@ -17,61 +14,71 @@ with lib;
...
-    pkgs.gcc
-    pkgs.smartmontools
-    pkgs.tcpdump
-    pkgs.gdb
-    pkgs.elfutils
-  ];
+  boot.kernel.sysctl = {
+    # reboot on kernel panic
+    "kernel.panic" = 60;
+    "kernel.panic_on_oops" = 1;
```
SuperSandro2000, line 24: In my experience this is not advisable at all and causes random reboots once every day or two.

build/common.nix
```diff
@@ -17,61 +14,71 @@ with lib;
...
 
-  boot.kernel.sysctl."kernel.panic" = 60;
-  boot.kernel.sysctl."kernel.panic_on_oops" = 1;
+  environment = {
+    enableDebugInfo = true;
+    systemPackages = with pkgs; [
+      # debugging
+      gdb
+      lsof
+      sqlite
```
SuperSandro2000, line 35:
```suggestion
      sqlite-interactive
```
otherwise the cli is not usable and I guess this is what we want here

build/common.nix
```diff
@@ -17,61 +14,71 @@ with lib;
...
-  nix.nrBuildUsers = 100;
+  services.openssh = {
+    enable = true;
+    authorizedKeysFiles = lib.mkForce [ "/etc/ssh/authorized_keys.d/%u" ];
+  };
 
   nix.extraOptions = ''
     allowed-impure-host-deps = /etc/protocols /etc/services /etc/nsswitch.conf
     allowed-uris = https://github.com/ https://git.savannah.gnu.org/ github:
   '';
```
SuperSandro2000, line 67:
```suggestion
  nix.extraOptions = ''
    allowed-impure-host-deps = /etc/protocols /etc/services /etc/nsswitch.conf
    allowed-uris = https://github.com/ https://git.savannah.gnu.org/ github:
  '';
```

build/common.nix
mweinelt: I have just refactored this, the infra has been running with this for years. And personally I cannot confirm such a behavior either.

PR #521, opened 2024-12-12
macs: add small wrapper to execute calls on all macs

mweinelt:
Adds a small wrapper script I use to execute commands on all macs in parallel.

### DIFF
--- a/macs/mac-exec
+++ b/macs/mac-exec
@@ -1,0 +1,16 @@
+#!/usr/bin/env bash
+
+HOSTS=(
+	"hetzner@enormous-catfish.mac.nixos.org"
+	"hetzner@growing-jennet.mac.nixos.org"
+	"hetzner@intense-heron.mac.nixos.org"
+	"hetzner@maximum-snail.mac.nixos.org"
+	"hetzner@sweeping-filly.mac.nixos.org"
+	"customer@eager-heisenberg.mac.nixos.org"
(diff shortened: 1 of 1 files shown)


macs/mac-exec
```diff
@@ -0,0 +1,16 @@
...
+	"hetzner@maximum-snail.mac.nixos.org"
+	"hetzner@sweeping-filly.mac.nixos.org"
+	"customer@eager-heisenberg.mac.nixos.org"
+	"customer@kind-lumiere.mac.nixos.org"
+)
+
+for host in "${HOSTS[@]}"; do
+	# shellcheck disable=SC2068
+	(ssh "${host}" -- $@ 2>&1| sed -e "s/^/${host} | /") &
+done
```
Mic92, line 16:
```
PIDS=()

for host in "${HOSTS[@]}"; do
    # shellcheck disable=SC2068
    (ssh "${host}" -- $@ 2>&1 | sed -e "s/^/${host} | /") &
    PIDS+=($!)
done

for pid in "${PIDS[@]}"; do
    wait $pid
done
```

vcunat: Maybe something like `wait "${PIDS[@]}"` at the end.

mweinelt: Applied in 63ea54bff9b83bea9a2307006defe17402da9c76

PR #524, opened 2024-12-23
builders: init profile based builder configuration

mweinelt:
With the following ideas:

- Use ZFS for RAID0
- Use tmpfs for /tmp and host build directories in RAM
- Run all builds in cgroups
- Updates and reboots using `system.autoUpgrade`
- Network autoconfiguration


mweinelt:
These are very much in production, so we can iterate on them further, but I'm merging them now, so we don't roll back any changes to hydras known_hosts or prometheus.

PR #532, opened 2025-01-06
hydra: increase max_output_size

mweinelt:
The new ISOs will have the LTS and latest kernel integrated as specialisations, which makes them about 205 MB larger on x86-64-linux.

What we get out of this is one less separate ISO for x86_64-linux and aarch64-linux we need to create and store.

### DIFF
--- a/build/hydra.nix
+++ b/build/hydra.nix
@@ -105,7 +105,7 @@ in
 
     allow_import_from_derivation = false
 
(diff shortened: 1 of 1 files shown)


ElvishJerricco (approved):

vcunat (approved):
I like the plan.

vcunat:
Deployed to mimas and restarted hydra-queue-runner.

PR #536, opened 2025-01-16
macs: configure cores/jobs per profile

mweinelt:
We need jobs with larger core counts to accomodate big build jobs, that would otherwise timeout.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -38,8 +38,7 @@ jobs:
     runs-on: macos-latest
     strategy:
       matrix:
-        # Doesn't seem that x86_64 is still in use?
-        machine: [arm64]
+        machine: [m1, m2-large]
     steps:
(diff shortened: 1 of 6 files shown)


macs/m1.nix
```diff
@@ -0,0 +1,8 @@
+{
+  # 8C, 16 GB, 256 GB
```
jfly, line 2: uber nit, feel free to ignore: What's the "256 GB"? Is that the size of the disk? Perhaps worth calling out explicitly? ("8 cores, 16 GB RAM, 256 GB storage")

jfly (approved):

Mic92 (approved):

macs/m1.nix
mweinelt: Yes, these are Macs. The reasoning is, that 256 GB RAM from Apple would bankrupt the foundation, so it cannot be that.

Issue #537, opened 2025-01-20
Cache diagnostic script broken


ttrssreal:
https://ix.io appears to have be retired 3:
Which breaks this script https://github.com/NixOS/infra/blob/6eecfb79e36ce3d4e14a71f36597b6f648cb2133/terraform/cache/diagnostic.sh
Not sure what paste-bin alternative is best

```
Pasted at: <!doctype html>
<html lang=en>
<head>
    <meta charset=utf-8>
    <title>&nbsp;</title>
</head>
<style>
    body {
        font-family: monospace;
	margin: 2em;
    }
</style>
<body>
    <p>ix.io is taking a break &#127867;</p>
    <img src="/underconstruction.gif" width="200px">
</body>
</html>
```

mweinelt:
It says they're under construction. What gives you the impression the service has been retired?

ttrssreal:
Maybe wrong wording; nonetheless, the script's broken.

mweinelt:
https://termbin.com is IMO a solid replacement

PR #539, opened 2025-01-21
Switch from `ix` to `termbin`

jfly:
This fixes https://github.com/NixOS/infra/issues/537

Demo:

```
$ terraform/cache/diagnostic.sh
...
> curl_test -6 https://cache.nixos.org/
Pasted at: https://termbin.com/8p7n

```

### DIFF
--- a/terraform/cache-staging/diagnostic.sh
+++ b/terraform/cache-staging/diagnostic.sh
@@ -1,5 +1,5 @@
 #!/usr/bin/env nix-shell
-#!nix-shell -i bash -p bind.dnsutils -p mtr -p curl
+#!nix-shell -i bash -p bind.dnsutils -p mtr -p curl -p netcat
 # shellcheck shell=bash
 # impure: needs ping
 #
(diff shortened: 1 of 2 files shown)


vcunat:
I'm not sure why I'm getting this warning, but the results seem OK.
```
> mtr -c 20 -w -r cache.nixos.org
/tmp/diagnostic.sh: line 33: warning: command substitution: ignored null byte in input
```

jfly:
Hmm... I don't seem to be able to reproduce that warning. Certainly strange that one of the commands we're running might print a 0 byte. Also seems fine to me for it to get ignored.

SuperSandro2000:
FYI termbin.com is not keeping files forever and they are eventually overwritten. Also It sometimes has race conditions.

jfly:
Do you have an alternative you'd suggest, @SuperSandro2000?

SuperSandro2000:
Unfortunately no.

Also I think on my work cooperate network it is actively being blocked.

jfly:
🤷 IMO, this is better than completely broken, which is what we currently have.

Perhaps it would be better to upload the report to the github issue itself? I suspect we could do that with the `gh` cli, and/or give people written instructions.

@mweinelt, what do you think?

vcunat:
I don't expect you can upload to GitHub without any login.  And that seems too cumbersome for a diagnostic script.

vcunat:
I agree that we should merge at least this.  Feel free to keep discuss and proposing improvements (e.g. a different location for the diags).

Mic92:
Don't think we need to keep diagnostic around forever anyway. Most issues are hopefully fixed eventually or should be re-tested later.

Issue #551, opened 2025-02-01
Address prometheus scraping jobs that are failing


jfly:
I see that we have 3 failing scraping jobs (at time of writing). [Query](https://prometheus.nixos.org/graph?g0.expr=up%20%3D%3D%200&g0.tab=1&g0.display_mode=lines&g0.show_exemplars=0&g0.range_input=1h):

- `up{instance="r13y.com:443", job="r13y"}`
- `up{instance="127.0.0.1:9190", job="rfc39"}`
- `up{instance="hydra.nixos.org:9199", job="hydra_notify"}`

# r13y

[Last successful scrape](https://prometheus.nixos.org/graph?g0.expr=up%7Bjob%3D%22r13y%22%7D%20%3D%3D%201&g0.tab=0&g0.display_mode=lines&g0.show_exemplars=0&g0.range_input=1y): 2024-09-21

Added in https://github.com/nixos/infra/commit/3c4f476eb63f4bf69ed0149e8852956c34a5f683.

Looks like it's a reproducibility checker created by @grahamc: https://github.com/grahamc/r13y.com.

Perhaps this can just be removed?

# rfc39

This is trickier. Seems to be periodically up ([link](https://prometheus.nixos.org/graph?g0.expr=up%7Bjob%3D%22rfc39%22%7D%20%3D%3D%201&g0.tab=0&g0.display_mode=lines&g0.show_exemplars=0&g0.range_input=8w)):Query

![Image](https://github.com/user-attachments/assets/bffa8d67-dcd6-4b15-a2c2-b966efb819e5)

Apparently this is a known "issue", see comment here: https://github.com/nixos/infra/blob/af0ed6d10dbb3a3ec919321314506b180d1f5faf/build/pluto/prometheus/exporters/rfc39.nix#L12.

AFAICT, we don't have any alerting rules configured that react to this (just [this systemd unit state](https://github.com/nixos/infra/blob/af0ed6d10dbb3a3ec919321314506b180d1f5faf/build/pluto/prometheus/exporters/rfc39.nix#L29)). Perhaps we could just stop scraping this? Is there useful historical data in here?

# hydra_notify

[Last successful scrape](https://prometheus.nixos.org/graph?g0.expr=up%7Bjob%3D%22hydra_notify%22%7D%20%3D%3D%201&g0.tab=0&g0.display_mode=lines&g0.show_exemplars=0&g0.range_input=1y): 2024-08-02

[...]

PR #552, opened 2025-02-01
Alert if any scraping job fails

jfly:
I'm marking this as a draft until https://github.com/NixOS/infra/issues/551 is resolved.

My understanding is that we really ought to have something like this to avoid a scenario where (for whatever reason: network issue, accidentally stopped exporter process, etc) prometheus is no longer successfully scraping targets.

Without this I believe many of our alerts would just happily continue to evaluate nothing at all.

### DIFF
--- a/build/pluto/prometheus/default.nix
+++ b/build/pluto/prometheus/default.nix
@@ -3,6 +3,7 @@
 {
   imports = [
     ./alertmanager.nix
+    ./exporters/up.nix
     ./exporters/blackbox.nix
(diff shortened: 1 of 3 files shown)


jfly:
Now that #551 is merged, this is ready for review.

Mic92:
Not yet deployed btw.

jfly:
Thanks, @Mic92! It looks like this must have gotten slurped up in a deploy, because I see it here: https://prometheus.nixos.org/alerts:

image

PR #558, opened 2025-02-16
Add a blackbox monitor of our mailserver

jfly:
To avoid potential alerting noise: I'll wait until this is deployed and succeeding before declaring an additional alert.

refs: https://github.com/NixOS/infra/issues/485
refs: https://github.com/Thaigersprint/thaigersprint-2025/issues/1

### DIFF
--- a/build/pluto/prometheus/exporters/blackbox.nix
+++ b/build/pluto/prometheus/exporters/blackbox.nix
@@ -1,7 +1,7 @@
-{ pkgs, ... }:
+{ config, pkgs, ... }:
 
 let
-  mkProbe = module: targets: {
+  mkStaticProbe = module: targets: {
     job_name = "blackbox-${module}";
     metrics_path = "/probe";
     params = {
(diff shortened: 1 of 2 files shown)


jfly:
Note: this is failing CI because of https://github.com/NixOS/infra/pull/559

build/pluto/prometheus/exporters/blackbox.nix
```diff
@@ -37,30 +12,104 @@ in
...
-        "https://status.nixos.org"
-        "https://survey.nixos.org"
-        "https://tarballs.nixos.org"
-        "https://weekly.nixos.org"
-        "https://wiki.nixos.org"
-        "https://www.nixos.org"
-        "https://tracker.security.nixos.org"
-      ])
+      {
+        job_name = "blackbox-https_success";
```
jfly, line 40: I opted to remove the existing `mkProbe` function, because I would have had to complicate it to handle `smtp_starttls` *and* `https_success`.

mweinelt: Yeah, we're not going to randomly make new mailservers appear, so I think `static_configs` is probably fine.

jfly: Sure, I can make that change when I have some time. Just to make sure I understand your reasoning, though: is the reason to keep things simple and keep the old `mkProbe` function around?

mweinelt: The relabeling config is what makes the configuration of the blackbox exporter scraping rule complicated, so I wouldn't want to repeat that for every probe. If you want to stick with dnssd create another function to keep the remaining config leaner.

jfly: I would prefer to use the MX record. I've renamed the old `mkProbe` to `mkHttpsProbe`, and added a new `mkSmtpStartTlsProbe`. Please take another look!

Mic92 (approved):
Can be merged if it works and has been deployed.

build/pluto/prometheus/exporters/blackbox.nix
mweinelt: mkStaticProbe, mkDNSSDProbe

jfly: Updated, PTAL.

mweinelt (approved):

PR #559, opened 2025-02-16
Remove unused variable `nix fmt` complains about

jfly:


mweinelt:
That won't eval. Pushed a fix in 470e1deb5666da4c01c222f735b48d6c25e26a26.

PR #562, opened 2025-02-21
fastly: implement "Lockable HTTP Tarball Protocol" (Flakes) for channels.nixos.org

emilylange:
This is part of another PR over in NixOS/nixos-channel-scripts that implements the precomputed x-amz-meta-link header. For the motivation behind this change look there (https://github.com/NixOS/nixos-channel-scripts/pull/76).

Hydra executes mirror-nixos-branch.pl when a given channel advances. Thus, we need some sort of fallback to handle channels which have long been EOL. A fallback also allows this change to be deployed in any order and over the span of multiple days.

Our default mode of operation is simply renaming the precomputed "x-amz-meta-link" header created by the script to "link", if it exists. Note that we cannot use "link" directly because AWS S3 does not allow it.

As a fallback we take the "location" header (which always exists) and template it into "link".
This lacks additional flake attributes like "rev" or whatever additional metadata the script may precompute, but is perfectly compliant with the "Lockable HTTP Tarball Protocol".

When running into the fallback, the string returned by nixpkgs' lib.trivial.versionSuffix will contain "dirty" instead of "pre-git" or the proper 7-char substring of the rev.

While we could have Fastly do a sub-request to fetch the git-revision txt right next to the tarball, I don't think it's worth the effort and complexity.

Tested using <https://fiddle.fastly.dev/>.

Ref: https://github.com/NixOS/nix/blob/61f49de7ae0b3899abdcc102832523153dd40d35/doc/manual/source/protocols/tarball-fetcher.md

### DIFF
--- a/terraform/channels.tf
+++ b/terraform/channels.tf
@@ -229,11 +229,24 @@ resource "fastly_service_vcl" "channels" {
       # Note: we need to match on 301s and 302s here, since Fastly has multiple
       # layers, and otherwise a redirect might still get cached at the second
       # layer after the first layer turned a 301 into a 302.
+      #
+      # Additionally, this also implements the "Lockable HTTP Tarball Protocol"
+      # to use nixexprs.tar.xz with Flakes and have it locked properly.
       if (beresp.status == 301 || beresp.status == 302) {
         set beresp.status = 302;
         set beresp.ttl = 0s;
         set beresp.grace = 0s;
         set beresp.cacheable = false;
+        if (req.backend.is_origin && std.suffixof(bereq.url, "/nixexprs.tar.xz")) {
+          # rename prepared link header if available
+          if (beresp.http.x-amz-meta-link) {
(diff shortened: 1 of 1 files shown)


terraform/channels.tf
```diff
@@ -229,11 +229,24 @@ resource "fastly_service_vcl" "channels" {
       # Note: we need to match on 301s and 302s here, since Fastly has multiple
       # layers, and otherwise a redirect might still get cached at the second
       # layer after the first layer turned a 301 into a 302.
+      #
+      # Additionally, this also implements the "Lockable HTTP Tarball Protocol"
+      # to use nixexprs.tar.xz with Flakes and have it locked properly.
```
Mic92, line 234: @emilylange do you still want to apply the changes you mentioned in matrix, or is this good as is?

emilylange: good as is :)

Mic92:
Tested in https://git.clan.lol/clan/clan-core/pulls/2921/files

PR #564, opened 2025-02-22
no longer depends on nixpkgs-unstable

Mic92:
the only place we were using this, was for the devshell.

But it's not really clear to me if we still need this?

cc @jfly

### DIFF
--- a/build/flake.nix
+++ b/build/flake.nix
@@ -98,16 +98,25 @@
           };
         }) self.nixosConfigurations;
 
-      # TODO: flake-utils.lib.eachDefaultSystem
-      devShell.x86_64-linux =
-        let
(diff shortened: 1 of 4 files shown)


jfly (approved):
LGTM.

For the record, [here's the reason I added it](https://github.com/NixOS/infra/pull/495#discussion_r1822026036):

> It's entirely non-obvious from this PR: I need sops's --filename-override flag, which was introduced in [sops 3.9.0](https://github.com/getsops/sops/blob/main/CHANGELOG.rst#390), which is not present in the stable nixos release we use here.

Which I see we have now:

```
$ gh pr checkout 564
$ direnv exec ./non-critical-infra sops --version
...
direnv: loading ...
sops 3.9.4 (latest)
```

PR #565, opened 2025-02-22
Build all core infra in CI

Mic92:
Depends on https://github.com/NixOS/infra/pull/564


.github/workflows/ci.yml
```diff
@@ -32,6 +32,9 @@ jobs:
           - caliban.nixos.org
           - elated-minsky
           - sleepy-brown
+          - haumea
```
jfly, line 35: Is the comment above (line 26) now vestigial?

> all builds combined consume too much disk space... we should soon switch to something else

(not a blocker) Could/should we be doing something else to generate this list of machines so we don't have to remember to keep it up to date as we add/remove/rename machines? Is that what https://github.com/nix-community/nix-github-actions is for?

jfly (approved):
This will be nice to have, thanks =)

Why did we have 2 flakes before? I had assumed it was about locking 2 different versions of `nixos-24.11`. Now our non-critical infra and critical infra will upgrade those together (which feels totally OK to me).

Mic92:
> This will be nice to have, thanks =)
> 
> Why did we have 2 flakes before? I had assumed it was about locking 2 different versions of `nixos-24.11`. Now our non-critical infra and critical infra will upgrade those together (which feels totally OK to me).

We could still add a second nixpkgs input if we need to and this approach would be still easier to manage I think.

.github/workflows/ci.yml
Mic92: nix-github-actions would also work. But the current solution works actually good enough and works for github forks. nix-fast-build will only rebuild machines if there are actually changes. So we might just keep it this way for simplicity.

jfly:
> We could still add a second nixpkgs input if we need to and this approach would be still easier to manage I think.

I totally agree a second input would be easier to manage.

I also don't think it's necessary, but I'll defer to more seasoned team members if they feel different.

.github/workflows/ci.yml
jfly: Sounds good. Feel free to mark this resolved once the comment on line 26 is updated (if it needs to be updated).

PR #570, opened 2025-02-24
Fix colmena

Mic92:


build/colmena.nix
```diff
@@ -0,0 +1,74 @@
...
+  options.colmena = {
+    hosts = lib.mkOption {
+      type = lib.types.attrsOf (
+        lib.types.submodule (
+          { name, ... }:
+          {
+            options = {
+              targetHost = lib.mkOption {
+                type = lib.types.str;
+                default = "${name}.nixos.org";
```
jfly, line 20: Do our machines have a fqdn configured that we could/should use instead?

jfly (approved):
LGTM!

build/colmena.nix
Mic92: Feels a bit like a layer violation to access the nixos module system from the flake parts module system. It might make parallel evaluation harder.

jfly: Sure. Perhaps this is a colmena question instead.

PR #580, opened 2025-03-10
dns: Add GitHub TXT verification for ngi.nixos.org

Erethon:
Follow up on #572. This is needed to verify the domain ownership on GitHub, sorry for not realizing sooner and having this as part of #572.


Mic92:
Applied.

PR #590, opened 2025-03-13
Alert if scraping jobs fail

jfly:
In addition to alerting on failing scraping jobs, we need to address our existing failures, which I researched in
https://github.com/NixOS/infra/issues/551. Copy pasting analysis from there:

- `r13y`: Last successful scrape: 2024-09-21. Website is down. @grahamc didn't respond to my query.
- `rfc39`: IMO, this is mis-architected. This scrapes a job that only runs periodically, which means we regularly get scrape failures. I've filed https://github.com/NixOS/rfc39/issues/14 with the upstream project seeking advice.
  - Since we don't actually alert on these metrics, I propose doing the simplest thing and just disabling this scrape job for now.
- `hydra_notify`: Last successful scrape: 2024-08-02. We disabled this service [in 2024](https://github.com/NixOS/infra/commit/66da5cfddb8c67b5e2f0b5bba62899f1f82eec42), there's no reason to keep scraping it.

(fixes https://github.com/NixOS/infra/issues/551)


Mic92:
applied

PR #592, opened 2025-03-15
Increase `for` for `DomainProbeFailure`

jfly:
Today we hit a stretch of over 3 hours where we could not successfully query `nix.ci`. Let's bump to a full day and see if we get that unlucky in the future.


mweinelt:
Applied.

Issue #586, opened 2025-03-13
Update mailing-lists with latest mailing lists and SMTP sending accounts in ImprovMX


jfly:
See [`mailing-lists`](https://github.com/NixOS/infra/blob/5eeb20e3a34afe3a1830357dc4d7b525d25f6282/non-critical-infra/modules/mailserver/default.nix#L48-L49). This needs to be updated with the latest data in ImprovMX.

This is worth scriptifying, because we'll have to do it again right before switching our MX records (in case anything changes in the mean time).

There are 2 parts to this:

- [x] https://github.com/NixOS/infra/pull/597: Import the mailing lists ("aliases" from ImprovMX. We can do this ourselves, just need a dump from ImprovMX.
  - To dump aliases from ImprovMX: `curl -X GET https://api.improvmx.com/v3/domains/nixos.org/aliases -H "Authorization: Basic api:$API_KEY" > /tmp/improvemx.json`
- [x] Set up STMP sending accounts. We cannot do this ourselves, because it requires passwords. I've emailed the 2 accounts (as documented on https://github.com/NixOS/org/blob/main/doc/mail.md). I'm waiting to hear back from them:
  - [ ] ~winter@nixos.org~: Skipped in order to not block the rollout. I've informed Winter, and she can always add herself when she does need access again.
  - [x] ngi@nixos.org: https://github.com/NixOS/infra/pull/600

jfly:
We discussed this in the infra matrix channel: we're going to call this done, even though @winterqt does not yet have a SMTP login account set up. I've informed her, and she can add herself whenever she needs it.

PR #597, opened 2025-03-22
Import ImprovMX aliases

jfly:
Note: there are 2 commits here: a refactor and then the import.
I suggest reading them independently.

This completes the aliases part of <https://github.com/NixOS/infra/issues/586>.

I wrote a script to parse a ImprovMX dump and generate this, but I don't want
to share it publicly because it has people's personal email addresses in
it. If you want to read the script, reach out to me on matrix, and I can
share it.


jfly:
Thanks, @Mic92! For the record, this is now deployed.

PR #601, opened 2025-03-29
dns: migrate to dnscontrol

mweinelt:
Purpose-built declarative and stateless DNS management that lets us overcome the limitation of the terraform API and the stale provider we have been using.

I'm not sure if the Github Actions part is a good idea and would love someone else's opinion. Especially the preview part seems as useful as it is risky.


.github/workflows/dns-apply.yml
```diff
@@ -0,0 +1,30 @@
+---
+name: Apply DNS changes
+
+on:
+  push:
+    branches:
+      - main
+    paths:
+      - dns/
```
jfly, line 9: Would this give me (a member of non-critical infra) power to deploy DNS changes (because I have the power to merge PRs)? Or would I not have the power to merge it because CODEOWNERS mentions `@NixOS/infra-build`? I suppose I could get my hands on the `secrets.NETLIFY_TOKEN` if I was being malicious.

.github/workflows/dns-preview.yml
```diff
@@ -0,0 +1,27 @@
...
+      - uses: actions/checkout@v4.2.2
+      - uses: cachix/install-nix-action@v31
+      - name: dnscontrol check
+        working-directory: ./dns/
+        run: |
+          nix run --inputs-from . nixpkgs#dnscontrol -- check
+      - name: dnscontrol push
+        if: github.repository == 'nixos/infra'
+        env:
+          NETLIFY_TOKEN: "${{ secrets.NETLIFY_TOKEN }}"
```
jfly, line 24: This looks like the same token as the deploy. Any chance we could use a read only token here instead?

formatter/flake-module.nix
```diff
@@ -16,7 +16,10 @@
 
         # older actionlint version don't recognize aarch64 builder
         programs.actionlint.enable = lib.versionAtLeast pkgs.actionlint.version "1.7.7";
-        programs.deno.enable = true;
+        programs.deno = {
+          enable = true;
+          excludes = [ "dns/*.js" ];
```
jfly, line 21: nit: why the exclude? Consider adding a quick comment explaining why.

dns/nixos.org.js
```diff
@@ -0,0 +1,186 @@
+D("nixos.org",
+	REG_NONE,
+	DnsProvider(DSP_NETLIFY),
+
+	//IGNORE("*", "NETLIFY"),
+	//IGNORE("*", "NETLIFY6"),
```
jfly, line 6: FYI these commented out lines of code are here.

dns/nixos.org.js
```diff
@@ -0,0 +1,186 @@
+D("nixos.org",
+	REG_NONE,
```
jfly, line 2: What's this about?

dns/nixos.org.js
```diff
@@ -0,0 +1,186 @@
...
+	CNAME("monitoring", "pluto.nixos.org."),
+	CNAME("prometheus", "pluto.nixos.org."),
+
+	// hydra builfarm
+	AAAA("eager-heisenberg.mac", "2a01:4f8:d1:a027::2"),
+	A("elated-minsky.builder", "167.235.95.99"),
+
+	AAAA("elated-minsky.builder", "2a01:4f8:2220:1b03::1"),
+	
+  A("enormous-catfish.mac", "142.132.140.199"),
```
jfly, line 80: nit: some funky indentation here

dns/nixos.org.js
jfly, line 1: There are 4 TODOs in `terraform/dns.tf` related to completing the rollout of the mailserver that I'd appreciate if we could port over to this file.

jfly (approved):
This looks nice to my eyes! Left a few comments inline, but overall LGTM.

I'm going to give you an approval, but happy to take another look if you want another review.

.github/workflows/dns-apply.yml
Mic92: It would be possible, but it would be also visible if you do it because, you cannot force push the malicious code afterwards.

Mic92: But I have a look if github's OIDC could be used to reduce the TTL of these token drastically.

Mic92: Doesn't look like external OIDC providers are supported by netlify.

.github/workflows/dns-preview.yml
mweinelt: No scoped token, which is why I consider netlify a bad provider.

dns/nixos.org.js
mweinelt: We don't interact with a registrar, so this is set to none.

dns/nixos.org.js
mweinelt: Ported.

jfly: Thank you!

.github/workflows/dns-apply.yml
winterqt: (Even if you could force push [do we not have branch protection set up here? {can infra-build members even override that?}], they're persisted in the audit log and [in the UI](https://github.com/NixOS/infra/activity) (which I didn't know about till recently!))

.github/workflows/dns-preview.yml
winterqt: For avoidance of doubt: at least GitHub Actions doesn't provide secrets to forked repos, which means the actual risk of this is really just constrained to those who can create branches here, e.g. non-critical infra and infra.

.github/workflows/dns-apply.yml
mweinelt: We require codeowner approval for merges (or repo admin bypass), so all DNS changes are controlled by in practice controlled by infra-build.

dns/nixos.org.js
```diff
@@ -0,0 +1,188 @@
...
+
+	TXT("_github-pages-challenge-ngi-nix.ngi", "4e8bffbb7ced2aec7be1f8cf3561d6"),
+	TXT("_github-pages-challenge-nixos", "f3a423ba6916e972cfb1e74f82f601"),
+
+	// netlify pages
+	CNAME("planet", "nixos-planet.netlify.app."),
+	CNAME("status", "nixos-status.netlify.app."),
+	CNAME("weekly", "nixos-weekly.netlify.com."),
+
+END);
```
winterqt, line 187: This is [no longer required](https://docs.dnscontrol.org/advanced-features/code-tricks#trailing-commas):

```suggestion
);
```

winterqt (commented):
I know I don't touch this repo nor am I a member of any infra team, but just as a general thought: maybe it'd be worthwhile to separate some of this configuration out into different, logically separated, files?

The giant file for the entire `nixos.org` zone is kind of unwieldy, so it may be useful to split it up so that things like Hydra and the builders are in one file, services infra is in another, NGI stuff is in another, wiki is in another, etc... (+/- any of these splits as you want, if you don't want it to be *too* split).

.github/workflows/dns-apply.yml
winterqt: > We require codeowner approval for merges (or repo admin bypass), so all DNS changes are controlled by in practice controlled by infra-build.

A theoretical evil version of @jfly could make a new branch here with a modified workflow that deploys any changes or leaks the Netlify token without codeowner approval (or even a PR), though, right?

dns/nixos.org.js
mweinelt: Ah wild, was added by the `get-zone` command on 4.14.3.

winterqt: Yeah, I noticed that it still does that; going to open a PR to fix it.

mweinelt:
> The giant file for the entire nixos.org zone is kind of unwieldy, so it may be useful to split it up so that things like Hydra and the builders are in one file, services infra is in another, NGI stuff is in another, wiki is in another, etc... (+/- any of these splits as you want, if you don't want it to be too split).

We are down to 185 lines from over 500 lines in terraform, I think we're good on this front, given that the zone now has lots of grouping and some comments.

What would you imagine splitting up to look like? I'm honestly fine with the complexity for this kind of domain.

dns/nixos.org.js
winterqt: nvm, already fixed (and I was looking at an old version): https://github.com/StackExchange/dnscontrol/releases/tag/v4.15.2

Bump your local dnscontrol? ;)

winterqt:
Yeah, good point. As I mentioned in that original comment I was thinking something like factoring out Hydra/the builders into their own file; the rest probably isn't required, but something like `ngi.js` so that the NGI team knows where their stuff is, `wiki.js` etc.

You're probably right though, it doesn't matter much given the line count and infrequency that we actually have to change DNS.

dns/nixos.org.js
mweinelt: It's whatever the shell has, so NixOS 24.11.

mweinelt:
Planning to merge in 12 hours as I think this change is uncontroversial.

.github/workflows/dns-preview.yml
```diff
@@ -0,0 +1,29 @@
...
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+    steps:
+      - uses: actions/checkout@v4.2.2
+      - uses: cachix/install-nix-action@v31
+      - name: dnscontrol check
+        working-directory: ./dns/
+        run: |
+          nix run --inputs-from . nixpkgs#dnscontrol -- check
```
Mic92, line 22: Does this need network? We could make this a static check as well.

mweinelt: It does not.

PR #603, opened 2025-03-30
non-critical-infra: set up debuginfo server

mweinelt:
Waiting for https://github.com/NixOS/nixpkgs/pull/394674


non-critical-infra/modules/debuginfod/index.html
```diff
@@ -0,0 +1,77 @@
...
+              alt="logo"
+            />
+          </a>
+        </p>
+
+        <p class="lead">
+          <code>https://debuginfod.nixos.org/</code> provides access to debug
+          symbols and derivation source code for debuggers supporting the
+          debuginfod protocol. Debug symbols are provided for all derivations
+          that enable the `separateDebugInfo` flag.
```
symphorien, line 54: this is not the case with this implementation of debuginfod (it only serves it for derivation present locally) but may be with the one proposed here https://github.com/NixOS/nixpkgs/pull/394674

non-critical-infra/modules/debuginfod/index.html
```diff
@@ -0,0 +1,77 @@
...
+        </p>
+        <p>
+          <pre>export DEBUGINFOD_URLS=http://debuginfod.nixos.org</pre>
```
symphorien, line 57:
```suggestion
          <pre>export DEBUGINFOD_URLS=https://debuginfod.nixos.org</pre>
```

non-critical-infra/modules/debuginfod/index.html
mweinelt: I wrote that whole sentence before you explained to me the limitations of what nixseparatedebuginfod can do. Would you mind proposing a change?

symphorien: it only serves debuginfo+source for derivations built with separateDebugInfo and whose .drv file is present on the machine running nixseparatedebuginfod. That's not very useful.

non-critical-infra/modules/debuginfod/default.nix
```diff
@@ -0,0 +1,23 @@
+{
+  imports = [
+    ../nginx.nix
+  ];
+
+  services.nixseparatedebuginfod.enable = true;
```
symphorien:
```suggestion
  services.nixseparatedebuginfod2 = {
    enable = true;
    substituters = [ "https://cache.nixos.org" ];
  };
```

I think nixseparatedebuginfod2 is not ready.
Substituters are explicitly set so that the local store is not served.

symphorien: I meant "now ready"...

PR #605, opened 2025-03-31
move dnscontrol check to nix

Mic92:


Mic92:
Seems to work fine and this will now also catch regressions in dnscontrol updates.

PR #613, opened 2025-04-06
Use zizmor to lint GitHub action workflows

mweinelt:
Inspired by https://github.com/NixOS/nixpkgs/pull/396451.

Our security posture needs to be at least as good as that of nixpkgs, if not better :wink:


mweinelt:
This is ready. PTAL

jfly (approved):
Neat tool! TIL about the artipacked attack :(

PR #616, opened 2025-04-06
dns: prepare Gandi migration

mweinelt:
ALIAS at the domain apex, CNAME for everything else.

### DIFF
--- a/.github/workflows/dns-apply.yml
+++ b/.github/workflows/dns-apply.yml
@@ -23,6 +23,7 @@ jobs:
       - uses: cachix/install-nix-action@d1ca217b388ee87b2507a9a93bf01368bde7cec2 # v31
       - name: dnscontrol push
         env:
+          GANDI_TOKEN: "${{ secrets.GANDI_TOKEN }}" # Expires 2026-04-07
           NETLIFY_TOKEN: "${{ secrets.NETLIFY_TOKEN }}"
         working-directory: ./dns/
         run: |
--- a/dns/nixos.org.js
+++ b/dns/nixos.org.js
@@ -173,7 +173,11 @@ D("nixos.org",
 	TXT("_github-pages-challenge-nixos", "f3a423ba6916e972cfb1e74f82f601"),
 
 	// netlify pages
+	ALIAS("@", "nixos-homepage.netlify.app."),
+	CNAME("common-styles", "nixos-common-styles.netlify.app."),
 	CNAME("planet", "nixos-planet.netlify.app."),
+	CNAME("search", "nixos-search.netlify.app."),
 	CNAME("status", "nixos-status.netlify.app."),
-	CNAME("weekly", "nixos-weekly.netlify.com.")
+	CNAME("weekly", "nixos-weekly.netlify.com."),
+	CNAME("www", "nixos-homepage.netlify.app.")
 );
--- a/dns/nixos.org.js
+++ b/dns/nixos.org.js
@@ -1,6 +1,6 @@
 D("nixos.org",
 	REG_NONE,
-	DnsProvider(DSP_NETLIFY),
+	DnsProvider(DSP_GANDI),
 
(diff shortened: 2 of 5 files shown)


.github/workflows/dns-apply.yml
```diff
@@ -23,6 +23,7 @@ jobs:
       - uses: cachix/install-nix-action@d1ca217b388ee87b2507a9a93bf01368bde7cec2 # v31
       - name: dnscontrol push
         env:
+          GANDI_TOKEN: "${{ secrets.GANDI_TOKEN }}"
```
Mic92, line 26: Would be useful to add the gandi expiration date here in case we forget to update it. Jonas said a year?

mweinelt: Comment added.

mweinelt:
Looks like it needs more changes on the Gandi end:

> ERROR: ERROR: error while getting Nameservers for zone="nixos.org" with provider="gandi": StatusCode: 404 ; Err: 404: Unknown domain

mweinelt:
Looks like we cannot sync the zone until we update the delegation. Meh!

image

.github/workflows/dns-apply.yml
```diff
@@ -23,6 +23,7 @@ jobs:
       - uses: cachix/install-nix-action@d1ca217b388ee87b2507a9a93bf01368bde7cec2 # v31
       - name: dnscontrol push
         env:
+          GANDI_TOKEN: "${{ secrets.GANDI_TOKEN }}" # Expires 2026-04-07
           NETLIFY_TOKEN: "${{ secrets.NETLIFY_TOKEN }}"
```
jfly, line 27: (I'm assuming it'll work find to deploy to netlify and gandi in parallel). Suggestion: consider adding a TODO/issue to remind us to clean up netlify (not just this code, but the old dns entries in netlify to avoid confusion).

dns/nixos.org.js
```diff
@@ -171,7 +172,11 @@ D("nixos.org",
 	TXT("_github-pages-challenge-nixos", "f3a423ba6916e972cfb1e74f82f601"),
 
 	// netlify pages
+	ALIAS("@", "nixos-homepage.netlify.app."),
```
jfly, line 175: Apologies if you explained this elsewhere: is this violating the advice from https://answers.netlify.com/t/support-guide-which-are-some-good-dns-providers-for-alias-aname-support/211?

> Some providers we cannot recommend after last review (which is well over 2 years ago at this point) are ALIAS or ANAME records provided by folks like Hover, DNS Made Easy, and even NS1 when not used as a service we provide: those “fancy” record types as implemented at those providers hosting was at last review incompatible with [the way we do GeoIP routing](https://www.ibm.com/products/ns1-connect) for our CDN.

Granted, this post from 2019 sounds a bit hand wavy about a "review" they did in 2017.

dns/nixos.org.js
```diff
@@ -171,7 +172,11 @@ D("nixos.org",
...
+	CNAME("common-styles", "nixos-common-styles.netlify.app."),
```
jfly, line 176: Are these CNAMEs (and the ALAIAS record above) going to apply cleanly to netlify (which, IIUC, we're still deploying to with the changes in this PR)?

dns/nixos.org.js
mweinelt: Yes, I explained that limitation in https://matrix.to/#/!qAqmHNDbGMBTaNCArk:nixos.org/$rOR7m5_eJbT8cJFHfChMLCS20ks0jOL0-c86qiwhqIA?via=nixos.org&via=matrix.org&via=lossy.network, and it is a reason we should be migrating away from netlify going forward.

dns/nixos.org.js
mweinelt: Sorry, there is no plan to apply this to Netlify at all. Will update the PR shortly.

jfly (approved):

mweinelt:
grafik

PR #622, opened 2025-04-09
fix(mailserver): enable Sender Rewriting Scheme (SRS) on umbriel

jfly:
We had deliverability issues when forwarding mail to other domains (if the domain of the email we were forwarding had a strict SPF policy, the receiving mailserver would drop it due to SPF fail). See https://github.com/NixOS/infra/issues/485#issuecomment-2787490679 for details.

PR #624, opened 2025-04-10
Roll out umbriel as our mailserver (again)

jfly:
Now that https://github.com/NixOS/infra/pull/622 has landed, I believe we're ready to try this again.

This reverts commit 5a6958d0002501486eb87714084c469369a78ac6.

CC @infinisil to confirm that things are working after this lands.


mweinelt:
Please instead revert the revert, which will show up with a "Reapply" prefix.

infinisil:
Can confirm that it works, thanks!

PR #632, opened 2025-04-10
[DO NOT MERGE UNTIL 2025-04-17] Use stricter SPF records

jfly:
For the record: I don't feel super strongly about this, and would be happy to close this unmerged. I wanted to file this to continue the discussion started on https://github.com/NixOS/infra/pull/509, when I was young and knew nothing about SRS.

Note: we still have 3 other `~all` SPF records (discourse, caliban, wiki). I imagine they could also get the same treatment if we want to be strict.

<hr>

This (effectively) reverts d65e24dd13b5336eac7f2cabdef577f9a7213fb5.

If you're sending emails to mailing lists that don't implement Sender Rewriting Scheme (SRS) then you might want to avoid using `-all`. IMO, this really is indicative of problem elsewhere. See https://gitlab.com/simple-nixos-mailserver/nixos-mailserver/-/issues/302 for details.

### DIFF
--- a/dns/nixos.org.js
+++ b/dns/nixos.org.js
@@ -13,7 +13,7 @@ D("nixos.org",
 		parts: [
 			"v=spf1",
 			"a:umbriel.nixos.org",
-			"~all"
+			"-all"
 		]
 	}),
 	DMARC_BUILDER({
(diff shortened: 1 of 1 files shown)


mweinelt:
I think this is the way to go.

jfly:
> I think this is the way to go.

In that case, I'm going to mark this ready for review. Out of an abundance of caution, I'm going to wait a least a week before merging this, because our mailserver is so new.

@mweinelt, what do we want to do with the other 3 SPF records we manage (discourse, caliban, wiki)?

mweinelt:
We should migrate all of them to `-all`.

jfly:
> We should migrate all of them to -all.

Great, updated.

mweinelt (approved):

Mic92 (approved):

PR #634, opened 2025-04-11
mailserver: enable arc signing

mweinelt:
Works similarly to DKIM and uses the same keys. Went for a 2048 bit RSA key for compat reasons. Larger ones are probably too large to put them into DNS.

Untested.


non-critical-infra/modules/mailserver/default.nix
```diff
@@ -41,6 +47,26 @@
...
+    # rspamadm dkim_keygen --selector arc-2025 --domain nixos.org --type rsa --bits 2048
+    sopsFile = ../../secrets/nixos.org-${arc.selector}-private-key.umbriel;
+    path = "/var/lib/rspamd/arc/nixos.org.${arc.selector}.key";
+  };
+
+  services.rspamd = {
+    overrides."arc.conf".text = ''
+      domain {
+        nixos.org {
+          selector = "${arc.selector}";
```
jfly, line 64: nit: unnecessary string interpolation?


```suggestion
          selector = arc.selector;
```

non-critical-infra/modules/mailserver/default.nix
```diff
@@ -41,6 +47,26 @@
     path = "${config.mailserver.dkimKeyDirectory}/nixos.org.mail.key";
   };
 
+  sops.secrets."nixos.org.${arc.selector}.key" = {
```
jfly, line 50: Should we reload `rspamd.service` when this changes? Please include [`restartUnits`](https://github.com/Mic92/sops-nix?tab=readme-ov-file#restartingreloading-systemd-units-on-secret-change) if appropriate.

dns/nixos.org.js
```diff
@@ -19,6 +19,7 @@ D("nixos.org",
 	DMARC_BUILDER({
 		policy: "none",
 	}),
+	TXT("arc-2025._domainkey", "v=DKIM1; k=rsa; p=MIIBIjANBgkqhkiG9w0BAQEFAAOCAQ8AMIIBCgKCAQEAwFK4Z4C0D3ea7Avv+oI25PS6WjcOh3A/2URanFtY6+oLpbtFdZi/Z/ou2VPTNcS1QUxw5pSmV4L1fcaVqM+elPHq1GN/38TkpwcZEnSHK5tX0PS5Jae2Q+e68yEZAwNDi5abjXswehuxX/F9R4GXpC/tOEyeHP8xfvRmDUq3mgIgWpfCcvdwQvxp25/umufjqGHdQXuq9/9yfpAL72hUXbOrvQ5hd56U7tv0/llIs5CyaGU76hA4kRXkd+iYUxcITzLjbc3ZRhIDs9b0zv0z2YPYgVgj2GiKL+TdcQ9z5Y5M0H4EGq6/Gn5fUILGRkCaBGvE3s3mY/aYUCvv4v0W9wIDAQAB"),
```
jfly, line 22: nit: Could we add a comment here helping people understand where the other half of this is? For example, see [this DKIM comment](https://github.com/nixos/infra/blob/e1b320825064a9f88c9cefdf908268f8636c2b53/dns/nixos.org.js#L120-L121).

non-critical-infra/modules/mailserver/default.nix
```diff
@@ -41,6 +47,26 @@
...
+    format = "binary";
+    owner = "rspamd";
+    group = "rpsamd";
+    mode = "0400";
+    # rspamadm dkim_keygen --selector arc-2025 --domain nixos.org --type rsa --bits 2048
```
jfly, line 55: nit: Could you elaborate on this a bit? Specifically where the public half of this comes from? See https://github.com/NixOS/infra/blob/d30eea236dc043be2b9c0852f17876fef65132e0/non-critical-infra/modules/mailserver/default.nix#L24-L36 for how I handled this for DKIM.

jfly (commented):
Thanks for tackling this! Code looks reasonable to me, I also have not tested it. I can deploy a version of this to my personal mailserver next week if that would be helpful.

Is ARC generally applicable for SNM users? Or is it really only useful if you're operating mailing lists and forwarding emails onto other domains? If it's generally applicable, I'd like to submit these instructions to [the SNM setup guide](https://nixos-mailserver.readthedocs.io/en/latest/setup-guide.html).

jfly (approved):
Forgot to click approve. LGTM

non-critical-infra/modules/mailserver/default.nix
Mic92: No this is a string. It's needed.

Mic92 (approved):

mweinelt:
> Is ARC generally applicable for SNM users? Or is it really only useful if you're operating mailing lists and forwarding emails onto other domains? If it's generally applicable, I'd like to submit these instructions to [the SNM setup guide](https://nixos-mailserver.readthedocs.io/en/latest/setup-guide.html).

Google and iCloud require it for bulk senders. Proposed in https://gitlab.com/simple-nixos-mailserver/nixos-mailserver/-/merge_requests/376.

non-critical-infra/modules/mailserver/default.nix
Mic92: Looks similar to my: https://github.com/Mic92/dotfiles/blob/main/machines/eve/modules/rspamd/rspamd.nix

jfly:
Discussed at today's infra meeting: @jfly to test this on his personal mailserver and then deploy this if all looks good

jfly:
@mweinelt, I deployed a [version of this](https://github.com/jfly/snow/compare/arc?expand=1) to my personal mailserver and while it didn't break anything, I couldn't find *any* evidence of this doing anything. Emails I sent directly as a SMTP-authenticated user, and emails I relay through it all don't show any ARC related headers when they land in my personal gmail. I also don't see anything showing up in rspamd logs.

mweinelt:
@Mic92 also has this configured, can you give that a try?

```
      allow_username_mismatch = true;

jfly:
I will try when I'm back at a computer!

jfly:
Adding `allow_username_mismatch = true;` doesn't see to have made a difference. I've been banging my head against this unproductively for too long. I'm not sure I'll have time to look at this again until next week, sorry!

PR #640, opened 2025-04-15
Add nixcon CFP email

infinisil:
Need an email for pretalx and we don't want to use the main one. Discussed in the NixCon org meeting from 2025-04-10

Ping @ra33it0, @ners, @Lassulus, @a-kenji

### DIFF
--- a/non-critical-infra/modules/mailserver/mailing-lists.nix
+++ b/non-critical-infra/modules/mailserver/mailing-lists.nix
@@ -124,6 +124,15 @@
       ];
     };
 
+    "cfp@nixcon.org" = {
+      forwardTo = [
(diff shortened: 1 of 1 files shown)


mweinelt:
We can adopt the nixcon.org for mail delivery in our mail setup later today or tomorrow.

Then we could go for cfp@nixcon.org, if that would be preferable.

infinisil:
I guess that would be better yeah! Then we should also use orgateam@nixcon.org, as has been the case in past yeah (and forward emails from nixcon@nixos.org to that one)

mweinelt:
The migration is done, please update the PR.

jfly (approved):

jfly:
This is now deployed.

PR #642, opened 2025-04-17
dns: migrate nixcon.org and set up mailing

mweinelt:
Migrates the DNS for nixcon.org to Gandi and management via dnscontrol through this repo.

Sets up mailing through umbriel.nixos.org and migrates some aliases, that are up for debate.

- orgateam@: primary contact address, currently nixcon@nixos.org
- foundation@ : does it make sense to still have that when the same name @nixos.org exists?
- volunteers@: unclear
- 2024@, 2024cfp@: are they still required? @Lassulus
- 2023@ was in Darmstadt and I can guarantee we don't require them any more
- 2022@ is even older, so I'm dropping it as well

Additionally, they all point to the same Google Group that I'm not sure gets used any more.

cc @infinisil

### DIFF
--- a/dns/nixcon.org.js
+++ b/dns/nixcon.org.js
@@ -1,0 +1,43 @@
+D("nixcon.org",
+	REG_NONE,
+	DnsProvider(DSP_GANDI),
+
+	MX("@", 10, "umbriel"),
+	SPF_BUILDER({
+		label: "@",
+		parts: [
+			"v=spf1",
+			"a:umbriel.nixos.org",
+			"-all"
+		]
+	}),
+	// Matching private key in `non-critical-infra/secrets/nixcon.org.mail.key.umbriel`
+	TXT("mail._domainkey", "p=MIGfMA0GCSqGSIb3DQEBAQUAA4GNADCBiQKBgQC1wQ2uPZfdlGmjDDxeNVet7IEFxS55TpWuqQWNKmd4fX8HcKKw7kVHXU5+gjT37wMUI27ZZnIobYhumnl+BLiXZqbuzAt7s3dbJU2de2ZWxOqcDRbK6m2A3AwIAiMzzRUjx14EWgnw55KRi2enpLyS0pKGdvSquHnxaySkAF8YIwIDAQAB"),
+	DMARC_BUILDER({
+		policy: "none",
+	}),
+
+	// Websites
+	TXT("_github-pages-challenge-nixcon", "6608e513e09036ab8cadb7ca4eb71b"),
+
+	ALIAS("@", "nixcon.github.io."),
+	CNAME("www", "nixcon.github.io."),
+
+	CNAME("2015", "nixcon.github.io."),
+	CNAME("2016", "nixcon.github.io."),
+	CNAME("2017", "nixcon.github.io."),
+	CNAME("2018", "nixcon.github.io."),
+	CNAME("2019", "nixcon.github.io."),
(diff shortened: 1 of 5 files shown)


Mic92 (approved):

non-critical-infra/modules/mailserver/mailing-lists.nix
```diff
@@ -8,6 +8,26 @@
   # an encrypted password. Run `nix run .#encrypt-email login -- --help` and
   # follow the instructions.
   mailing-lists = {
+    # nixcon.org
+    "orgateam@nixcon.org" = {
+      forwardTo = [
+        ../../secrets/nixcon-list-email-address.umbriel
```
jfly, line 14: FYI, we don't have to encrypt these. I personally would be inclined to just hard-code "nixcon@nixos.org" here instead.

jfly (approved):
LGTM. A couple comments inline that I trust you to resolve as you see fit.

Lassulus:
No. The old aliases are not necessary anymore.

non-critical-infra/modules/mailserver/mailing-lists.nix
infinisil: This alias we should keep, it's linked on https://github.com/NixCon (which we can change, but I imagine some people have that stored, and it might be linked elsewhere too)

non-critical-infra/modules/mailserver/mailing-lists.nix
```diff
@@ -8,6 +8,26 @@
...
+      ];
+    };
+
+    "volunteers@nixcon.org" = {
+      forwardTo = [
+        ../../secrets/nixcon-list-email-address.umbriel
+      ];
+    };
```
infinisil, line 28: These ones seem removable. If you do keep the foundation one, should probably forward to `foundation@nixos.org` though.

mweinelt: We should probably only keep what we still reference in the current year.

non-critical-infra/modules/mailserver/mailing-lists.nix
mweinelt: Up until now this is a different recipient. A google group.

mweinelt: Replaced with a forward to nixcon@nixos.org and dropped the Google group entirely.

non-critical-infra/modules/mailserver/mailing-lists.nix
mweinelt: Dropped both.

mweinelt: Confirmed with zimbatm that neither Pretix nor Pretalx use any of these addresses.

dns/nixcon.org.js
```diff
@@ -0,0 +1,43 @@
+D("nixcon.org",
+	REG_NONE,
+	DnsProvider(DSP_GANDI),
+
+	MX("@", 10, "umbriel"),
```
riotbib, line 5: The MX record set here translates to `umbriel.nixcon.org` which is non-existing currently. Should it maybe be `umbriel.nixos.org`?

vcunat: Yes, it seems clear.  Thanks!  https://github.com/NixOS/infra/pull/647

vcunat: Deployed.

Issue #626, opened 2025-04-10
Evaluate anubis for hydra.nixos.org


mweinelt:
**Is your feature request related to a problem? Please describe.**

The `hydra-server` component is pretty fragile when hit with many expensive requests and previously required manual restarts to become available again.

To that end we've previously implemented abb78774c02935d9e3e90059fcbba4b73d39717b and c7a93adf21b76de9d3d1e721b4aa4d39fa99f4d9.

**Describe the solution you'd like**

Evaluate anubis as a proof-of-work (hashcash) scheme that everyone claiming to be an actual browser needs to solve.

**Describe alternatives you've considered**

Continue evaluating and updating the banned user agents list.

**Additional context**

Anubis was packaged into nixos-unstable and will be available in 25.05. https://github.com/NixOS/nixpkgs/pull/392018

Currently, ever 4th-5th requests gets blocked based on the user agent.

```
[root@mimas:/var/log/nginx]# wc -l abuse.log.1
902154 abuse.log.1

[root@mimas:/var/log/nginx]# wc -l access.log.1
2958883 access.log.1
```

The ratio has been much worse for earlier periods. E.g. in Feb 2025 we blocked 7 times as many requests as we accepted.

SuperSandro2000:
I would recommend to wait/implement proxy_auth support in anubis and not proxy everything through it. see https://github.com/Xe/x/pull/694

mweinelt:
I think that can be an incremental improvement. I agree that we want it, because piping full ISOs through Anubis is likely not ideal.

mweinelt:
https://git.gammaspectra.live/git/go-away#why

PR #663, opened 2025-04-24
hydra-proxy: replace abuse handling with anubis

mweinelt:
Closes: #626

### DIFF
--- a/build/hydra-proxy.nix
+++ b/build/hydra-proxy.nix
@@ -1,32 +1,25 @@
 {
   config,
-  lib,
   pkgs,
   ...
 }:
 
-let
-  bannedUserAgentPatterns = [
-    "Trident/"
-    "Android\\s[123456789]\\."
-    "iPod"
-    "iPad\\sOS\\s"
-    "iPhone\\sOS\\s[23456789]"
-    "Opera/[89]"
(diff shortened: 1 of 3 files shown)


jfly (approved):
LGTM!

style nit, feel free to ignore: I personally like using nix to inject `config.services...` references to port numbers rather than hardcoding the port number everywhere.

build/hydra-proxy.nix
```diff
@@ -1,33 +1,26 @@
...
-    "Windows\\s95"
-    "Windows\\s98"
-    "Windows\\sNT\\s[12345]\\."
-  ];
-in
 {
   networking.firewall.allowedTCPPorts = [
     80
     443
+    9001
```
SuperSandro2000, line 11: Shouldn't we limit that to Prometheus and not make it public?

mweinelt: Eventually, but up until now all exporters have been publically accessible.

SuperSandro2000: That is not that great.

Depending on the exporter this might be more or less of a problem. Some just expose internal counters but others execute commands on scrape and I am not sure if they have rate limit and if you could cause damage by spamming those endpoints.

PR #666, opened 2025-04-29
Move away from deprecated in-line `lifecycle_rule` and `cors_rule`

arianvp:
AWS terraform provider deprecated the use of `lifecycle_rule` and `cors_rule`.

TODO: I'm not 100% sure if this applies without "Downtime". There might be a brief few seconds where the CORS rule is not set. Is that a problem?

### DIFF
--- a/terraform/cache.tf
+++ b/terraform/cache.tf
@@ -5,15 +5,26 @@ locals {
 resource "aws_s3_bucket" "cache" {
   provider = aws.us
   bucket   = "nix-cache"
+}
 
-  lifecycle_rule {
-    enabled = true
+resource "aws_s3_bucket_lifecycle_configuration" "cache" {
+  provider = aws.us
+  bucket   = aws_s3_bucket.cache.id
+
+  rule {
+    id     = "transition-to-standard-ia"
+    status = "Enabled"
 
     transition {
       days          = 365
       storage_class = "STANDARD_IA"
     }
   }
+}
+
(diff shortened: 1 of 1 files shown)


mweinelt:
IMO: We won't let brief outages stop us from fixing technical debt.

terraform/cache.tf
```diff
@@ -5,15 +5,26 @@ locals {
...
       storage_class = "STANDARD_IA"
     }
   }
+}
+
+resource "aws_s3_bucket_cors_configuration" "cache" {
+  provider = aws.us
+  bucket   = aws_s3_bucket.cache.id
 
   cors_rule {
```
jfly, line 29: Did you intend to leave `cors_rule` here?

arianvp: yes. that is the syntax

https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/s3_bucket_cors_configuration

jfly: Ah, sorry. I just saw "deprecated [...] `cores_rule`" in the title, and missed the important "in-line" part. Please disregard.

mweinelt (approved):

Mic92:
@arianvp I put the other pull request that is blocking you in the merge queue. So you can rebase afterwards.

arianvp:
Awesome. I just found out some mistakes in this PR that I have to clean up first. But at least I can `plan` without seeing drift which will help with iteration

PR #682, opened 2025-05-13
Add info about the team meeting

jfly:
I now have access to the nixos.org Google Calendar. Hopefully this will help with off-by-one confusion about which Thursday is the day we meet.

I also believe the time was actually wrong before: AFAIK, there is no place on earth that observes CET year round (Zurich switches between CET and CEST).

Full disclosure: I basically copied this wording from the [nix formatting team
readme](https://github.com/NixOS/nixfmt/blob/master/team/README.md).


jfly:
I'm going to take that as a shipit =) Feel free to pick a different name for the timezone!

PR #687, opened 2025-05-16
channels: Add NixOS 25.05 as beta

leona-ya:
Part of the release process.

Please wait for the 25.05 jobsets to be created tho.

### DIFF
--- a/channels.nix
+++ b/channels.nix
@@ -39,6 +39,22 @@ rec {
       status = "rolling";
     };
 
+    "nixos-25.05" = {
+      job = "nixos/release-25.05/tested";
+      variant = "primary";
+      status = "beta";
+    };
+    "nixos-25.05-small" = {
+      job = "nixos/release-25.05-small/tested";
+      variant = "small";
+      status = "beta";
+    };
+    "nixpkgs-25.05-darwin" = {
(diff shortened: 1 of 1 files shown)


vcunat:
The three jobsets are there.  Note that our scripts will keep failing with an alert if this is merged before all of them have at least one evaluation with all builds finished.  (not a big deal; I don't know how urgent it is to get a -small channel now, for example)

channels.nix
```diff
@@ -39,6 +39,22 @@ rec {
...
+      job = "nixos/release-25.05-small/tested";
+      variant = "small";
+      status = "beta";
+    };
+    "nixpkgs-25.05-darwin" = {
+      job = "nixpkgs/nixpkgs-25.05-darwin/darwin-tested";
+      variant = "darwin";
+      status = "beta";
+    };
+
```
Jayman2000, line 57: I tried running `nix --extra-experimental-features 'nix-command flakes' flake check --all-systems` in order to check this PR. It gave me an error about trailing whitespace. I was able to fix the error by making this change:

```suggestion
    };

```

Jayman2000 (changes requested):

channels.nix
mweinelt: Run `nix fmt`.

PR #699, opened 2025-05-28
send fundraising emails also to foundation@nixos.org

Lassulus:


infinisil:
Consensus among the board that this is good, so this can be merged :)

PR #704, opened 2025-05-31
mailinglists: remove community@nixos.org

Lassulus:
We only get spam send to this mailing list, or requests the moderation team can't handle. So better just remove it.

Maybe we also need to remove it from the website somewhere if it is listed there.

### DIFF
--- a/non-critical-infra/modules/mailserver/mailing-lists.nix
+++ b/non-critical-infra/modules/mailserver/mailing-lists.nix
@@ -22,12 +22,6 @@
       ];
     };
 
-    "community@nixos.org" = {
-      forwardTo = [
-        "moderation@nixos.org"
-      ];
-    };
-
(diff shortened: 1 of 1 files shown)


jfly (changes requested):
> Maybe we also need to remove it from the website somewhere if it is listed there.

Yeah, I see it listed here: https://github.com/NixOS/nixos-homepage/blob/main/core/src/pages/community.astro#L220 (https://nixos.org/community/). Once that's resolved, we can deploy this.

Lassulus:
after some delays: https://github.com/NixOS/nixos-homepage/pull/1784

jfly:
Now that https://github.com/NixOS/nixos-homepage/pull/1784 has landed, we're good to deploy this.

jfly (approved):
Oops, forgot to "approve".

Issue #700, opened 2025-05-28
Shared team email inboxes with Freescout


infinisil:
**Is your feature request related to a problem? Please describe.**
Currently, `@nixos.org` emails for teams get [forwarded](https://github.com/NixOS/infra/blob/8778e8512d7977bc260c1ffcd7cd4681521697b3/non-critical-infra/modules/mailserver/mailing-lists.nix) to personal emails. While this is simple, it makes managing emails tricky:
- Everybody needs to locally maintain the list of emails that still need attention, wasting a lot of our attention.
- Notifications for all emails are distracting when it doesn't pertain to most people.
- There is no way to easily delegate or synchronise on responses.

Overall, a lot of effort is duplicated and wasted, and we're really noticing this in the foundation board.

**Describe the solution you'd like**

Looking into potential solutions, I've found [Freescout](https://freescout.net/) to be ideal. It has to be backed by a mail server, but has its own web UI on top with exactly the features we need:
- A way to assign people to emails and mark threads as complete
- Configure notifications and other automations
- _Much_ more

In addition to being open-source, Freescout has an interesting monetization model: The base product is free, but any [extension modules](https://freescout.net/modules/) (of which there are many neat ones!) cost, but it's only a one-time fee, no subscription!

On behalf of the @NixOS/foundation board, I'd like this to be deployed for foundation@nixos.org, which means:
- To stop foundation@nixos.org from forwarding emails and instead have it serve as a standard inbox with a SMTP and IMAP login.
- To deploy Freescout on https://freescout.nixos.org. We can use this great recent effort to package Freescout by @e1mo: https://cyberchaos.dev/e1mo/freescout-nix-flake
- To make me a Freescout administrator, so I can configure it with the foundation's inbox, invite the other board members and register extension modules.

**Describe alternatives you've considered**
There's hacky ways of having shared read markers on emails, but that leaves much unaddressed. 

**Additional context**

Note that with this, emails (which are sometimes private) will be stored on the infra teams servers going forward, and any Freescout admins will also have access to all configured email inboxes, which makes emails less private than they are now. We expect such access to be reasonably restricted, and those who have it to maintain confidentiality.

We're happy to try this with the foundation email and see how it goes, but expect other teams to want to join in too if it works well.

While it would be great if the infra team could implement this for us, I'm also happy to jump in if some extra hands are necessary!

Ping @jfly @mweinelt (as you are very involved in email handling recently).

SuperSandro2000:
[...]

PR #708, opened 2025-06-02
Add freescout cname to umbriel

jfly:
This is necessary to test out freescout. https://github.com/NixOS/infra/issues/700


vcunat (approved):
OK, why not.

PR #715, opened 2025-06-03
Add support for storing emails in mailboxes accessible via IMAP

jfly:
This is a prerequisite for setting up Freescout. Note: this means we need to think about backups for umbriel, which is tracked in https://github.com/NixOS/infra/issues/700.


infinisil (approved):
Awesome, thanks a lot for getting this working!

PR #716, opened 2025-06-03
terraform-iam: tune archeologist permissions

zimbatm:
The main change is to give them read-only access to the main s3://nix-cache/ on top of the side ones.

The profile had two sid with overlapping permissions, so I removed one side.

Finally, we also given them access to CloudWatch metrics so they can see the bucket sizes.


zimbatm:
Since nobody is using the archeologist profile, I applied the change directly. I had to do a targeted apply because it was trying to delete other resources.

mweinelt (approved):

PR #717, opened 2025-06-05
add gh-releases.nixos.org

Mic92:
This is a domain that can be used for hosting releases from GitHub.
We use our own CDN to provide IPv6 support and retain the flexibility
to move to a different provider in the future if necessary.

We might integrate this into releases.nixos.org in the future, but
for now this serves as a testing ground.


Mic92:
This is a preparation for https://github.com/NixOS/experimental-nix-installer

cc @tomberek @Lassulus @abathur @mkenigs

PR #718, opened 2025-06-05
Fastly configuration for gh-releases.nixos.org

Mic92:
• Add Fastly CDN service for gh-releases.nixos.org to provide friendly URLs for GitHub releases

URL mapping examples:
  - gh-releases.nixos.org/nix/0.27.0/
nix-installer.sh  → github.com/NixOS/experimental-nix-installer/releases/download/0.27.0/
nix-installer.sh
  - gh-releases.nixos.org/patchelf/0.18.0/patchelf-x86_64-linux → github.com/NixOS/patchelf/releases/download/0.18.0/patchelf-x86_64-linux

### DIFF
--- a/terraform/gh-releases.tf
+++ b/terraform/gh-releases.tf
@@ -1,0 +1,97 @@
+locals {
+  gh_releases_domain = "gh-releases.nixos.org"
+}
+
+resource "fastly_service_vcl" "gh_releases" {
+  name        = local.gh_releases_domain
+  default_ttl = 3600
+
+  backend {
+    address               = "github.com"
+    auto_loadbalance      = false
+    between_bytes_timeout = 10000
+    connect_timeout       = 1000
+    error_threshold       = 0
+    first_byte_timeout    = 15000
+    max_conn              = 200
+    name                  = "github.com"
+    override_host         = "github.com"
+    port                  = 443
+    ssl_cert_hostname     = "github.com"
+    ssl_check_cert        = true
+    use_ssl               = true
+    weight                = 100
+  }
+
+  request_setting {
+    name      = "Redirect HTTP to HTTPS"
+    force_ssl = true
+  }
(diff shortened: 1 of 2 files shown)


terraform/gh-releases.tf
```diff
@@ -0,0 +1,185 @@
...
+    period            = local.fastlylogs["period"]
+    message_type      = "blank"
+    s3_iam_role       = local.fastlylogs["iam_role_arn"]
+  }
+}
+
+resource "fastly_tls_subscription" "gh_releases" {
+  domains               = [for domain in fastly_service_vcl.gh_releases.domain : domain.name]
+  configuration_id      = local.fastly_tls12_sni_configuration_id
+  certificate_authority = "globalsign"
```
mweinelt, line 180:
```suggestion
  certificate_authority = "lets-encrypt"
```

infinisil:
Hmm, shouldn't we maintain the "experimental" label? It's really easy for people to think it's stable if it's not explicitly marked as experimental

Mic92:
> Hmm, shouldn't we maintain the "experimental" label? It's really easy for people to think it's stable if it's not explicitly marked as experimental

I would like to not have this as part of the URL but rather at the place where link to it. Otherwise we would have to migrate it again, when we declare it stable, which is another friction point.

Mic92:
> You picked this name so it would be stable if we switched the backend around. But now you prefixed the hostname with `gh-`, which seems off. Wondering if we should invest time to brainstorm another name. I understand the difficulty being that releases.nixos.org would have been ag ood fit but is already taken.

We could potentially move all releases to github as well.

Mic92:
curl -v https://artifiacts.nixos.org/nix/0.27.0/nix-installer.sh is the new URL: https://github.com/NixOS/infra/pull/751

PR #723, opened 2025-06-09
draupnir: migrate from mjolnir

mweinelt:


emilylange:
It appears as if you forgot to delete https://github.com/NixOS/infra/blob/aa43b31b3f5c20da1200aa7d5c664b303245f2e2/non-critical-infra/secrets/mjolnir-password.caliban :face_with_peeking_eye:

mweinelt:
It's the only place where the password is documented, and I still have a note to migrate it to bitwarden.

PR #745, opened 2025-06-16
Store email and allow login for finance@nixos.org

infinisil:
Allows onboarding onto freescout, just like foundation@

Will set it up to be accessible by all foundation members and Kate


jfly:
This PR is deployed.

PR #746, opened 2025-06-16
Store NGI email

infinisil:
@fricklerhandwerk asked me to also onboard the ngi email onto freescout, so the mails needs to be stored on the server.

Note that to be able to onboard everybody onto freescout, I'll also need the emails it currently gets forwarded to and the login credentials.


jfly:
This PR is deployed.

> I will DM you this later today over matrix.

This is done.

PR #747, opened 2025-06-18
remove erictapen for ngi@nixos.org in mailing-lists.nix

erictapen:
If I understand this right from @infinisil's email, I should access this address through freescout now.


infinisil:
There's no "should", currently both work, and I figured you might want to try out Freescout for a bit before turning off the personal forwards (currently both work), but entirely up to you :)

PR #748, opened 2025-06-18
Store moderation team email for freescout

infinisil:
@Lassulus asked me to onboard the @NixOS/moderation team onto freescout, so this PR sets the groundwork for that.

The password I generated won't be needed after I set up freescout, but I shared it with @Lassulus, who put it in the moderation teams vault collection.


Lassulus (approved):

mweinelt:
The change is now deployed.

PR #752, opened 2025-06-20
Alias cfp.nixcon.org to talks.nixcon.org

infinisil:
See https://matrix.to/#/!GlDwnQuDPMNzaUXCJB:lassul.us/$XlMOVN5dZnWuL6f-bidD-yV1aHtJC57uqOtqp1Z_hFs?via=lassul.us&via=matrix.org&via=nixos.dev

cfp.nixcon.org isn't valid anymore

Ping @Lassulus

### DIFF
--- a/dns/nixcon.org.js
+++ b/dns/nixcon.org.js
@@ -36,7 +36,7 @@ D("nixcon.org",
 	CNAME("2025", "nixcon.github.io."),
 
 	// Scheduling
-	CNAME("cfp", "pretalx.com."),
+	CNAME("cfp", "talks.nixcon.org."),
 	CNAME("talks", "pretalx.com."),
 
 	// Ticketing


mweinelt:
How is that supposd to work? This removes just an indirection in DNS, that doesn't make cfp.nixcon.org work, because the SNI and Host header will still request that vhost.

infinisil:
Oh I guess an `ALIAS` record should work then, although I'm not an expert. Alternatively @Lassulus should be able to just have both `cfp` and `talks` in pretalx

mweinelt:
What you probably need is an HTTP redirect. Or we drop the domain entirely.

mweinelt:
> https://github.com/NixOS/infra/pull/752

Same issue. `ALIAS` is just a custom record to support a CNAME-like construct at the domain apex.

PR #756, opened 2025-06-26
Store hardware team emails to be able to set up freescout

infinisil:
@Mic92 requested the hardware team to use freescout too.

I'll send the email password to Mic92 to put into the vault, though not strictly necessary as it's set-and-forget in freescout.

Same as #748 and #746


mweinelt (approved):

mweinelt:
Deployed.

PR #763, opened 2025-07-04
Scrape Postgres and custom SQL queries from tracker.security.nixos.org

Erethon:
This is needed for https://grafana.nixos.org/d/beo2uotj65lvkb/nix-security-tracker?orgId=1&from=now-6h&to=now&timezone=browser&var-Instance to work.

I renamed the postgresql job scraper from `haumea-postgresql` to `postgresql` since it's a generic postgresql scraping job. I looked at the dashboards in Grafana and I think we didn't hardcode the old job name anywhere, so this shouldn't break anything.


mweinelt:
Deployed.

PR #764, opened 2025-07-08
builders: prep for build dir move in nix 2.30

mweinelt:
>    build-dir no longer defaults to $TMPDIR
>
>    The directory in which temporary build directories are created no longer defaults to TMPDIR or /tmp, to avoid builders making their directories world-accessible. This behavior allowed escaping the build sandbox and can cause build impurities even when not used maliciously. We now default to builds in NIX_STATE_DIR (which is /nix/var/nix/builds in the default configuration).

via https://discourse.nixos.org/t/nix-2-30-0-released/66449

### DIFF
--- a/builders/common/nix.nix
+++ b/builders/common/nix.nix
@@ -6,6 +6,17 @@
 }:
 
 {
+  fileSystems."/nix/var/nix/builds" = {
+    device = "none";
+    fsType = "tmpfs";
+    options = [
+      "huge=within_size"
+      "mode=0700"
+      "nosuid"
+      "nodev"
+    ];
+  };
+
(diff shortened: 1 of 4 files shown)


vcunat:
Perhaps it's worth keeping `/tmp` on tmpfs in addition to this?  (though it surely doesn't need to be large)

Mic92:
In our nixos infra we could have our own custom build-dir mounted set to /tmp2 if we wanted. But a better upstream default would be still desired for those that want to mass-rebuild on their own machines.

Mic92:
https://github.com/NixOS/nix/pull/13839 was merged.

PR #765, opened 2025-07-09
Add @gefla to the nixcon mailing list

infinisil:
Now part of the organisers


jfly (approved):

vcunat:
I guess it depends which machines, but I didn't know our configs needed heavy building.  I'd expect almost every nontrivial derivation gets pulled from cache.nixos.org.

PR #779, opened 2025-07-17
hydra-proxy: limit per client request rate

mweinelt:
Put a rate limit on per source adress requests to hydra-server

> Excessive requests are delayed until their number exceeds the maximum burst size in which case the request is terminated with an [error](https://nginx.org/en/docs/http/ngx_http_limit_req_module.html#limit_req_status).

https://nginx.org/en/docs/http/ngx_http_limit_req_module.html

### DIFF
--- a/build/hydra-proxy.nix
+++ b/build/hydra-proxy.nix
@@ -43,6 +43,9 @@
         default "anubis";
         nix.dev-Uogho3gi "hydra-server";
       }
+
+      limit_req_zone $binary_remote_addr zone=hydra-server:8m rate=2r/s;
+      limit_req_status 429;
     '';
 
     eventsConfig = ''
@@ -77,6 +80,9 @@
 
       locations."/" = {
         proxyPass = "http://$upstream";
+        extraConfig = ''
+          limit_req zone=hydra-server burst=5;
+        '';
       };
 
       locations."~ ^/build/\\d+/download/" = {


SuperSandro2000:
How did you come up with those numbers?

On first page load my browser did 3 requests outside the static/ directory to hydra.nixos.org. I think the proposed rate limit is very tight and I think with normal user navigation you could easily hit this.

Also this makes the static section a few lines below mandatory and part of the rate limit concept as otherwise you would be rate limited on first page load.

mweinelt:
Feel free to try it out and tell me if you can hit the ratelimit. It has been deployed since before I opened this PR for test purposes.

SuperSandro2000:
Hmmm, it is configured with delay instead of returning 429 which makes it a bit hard to differentiate if I am hitting the rate limit or hydra is just slow as usual. 

But I think I've hit the rate limit after some light weight browsing because pages like https://hydra.nixos.org/job/nixpkgs/staging/stdenv.aarch64-darwin suddenly took 9.5s to load.

mweinelt:
Because you don't see what I see, here are the aggregated access numbers in the last 15 minutes:

```
[root@mimas:/var/log/nginx]# grep "18/Jul/2025:13" access.log | grep -v /static -c
118825

[root@mimas:/var/log/nginx]# grep "18/Jul/2025:13" access.log | grep -v /static | grep " 200 " -c
104855

[root@mimas:/var/log/nginx]# grep "18/Jul/2025:13" access.log | grep -v /static | grep " 503 " -c
5143
```

This is like 132 requests per second, so I think you can't even differentiate between being delayed or whether the machine is currently very loaded and needs more time to serve expensive requests.

PR #805, opened 2025-08-15
Remove personal nixcon email forwards

infinisil:
Also alias cfp to nixcon, because we can use freescout to manage that after #800 

Ping @ners @ra33it0 @Lassulus @a-kenji because of the cfp@nixos.org change


jfly (approved):
LGTM, but should I wait before merging/deploying?

infinisil:
No need to wait, I take the blame if people were to complain ;)

PR #811, opened 2025-08-25
mimas: migrate to nftable and configure whole as blocks

mweinelt:


vcunat:
1. The top comment shows expiration after 6 hours?  (but no idea how that works really)

jfly:
> It prints the error to stdout/stderr and moves on. Not silly at all.

That's not because you're setting `check=True`, it's because the subprocess is inheriting the parent process's stdout/stderr. For example:

<details>

<summary>demo.py</summary>

```python
#!/usr/bin/env python

import subprocess

noisy_error = (
    "echo 'some info' >/dev/stdout && echo 'this is an error' >/dev/stderr && exit 42"
)

print("=== Experiment 1: Running `noisy_error` with `check=False`")
subprocess.run(noisy_error, shell=True, check=False)
print("=" * 30)

print(
    "=== Experiment 2: Running `noisy_error` with `check=True`, and catching and tossing the Python exception"
)
try:
    subprocess.run(noisy_error, shell=True, check=True)
except subprocess.SubprocessError:
    pass
print("=" * 30)
```
</details>

Note how both Experiment 1 and 2 output the exact same thing:
```console
$ python demo.py
=== Experiment 1: Running `noisy_error` with `check=False`
some info
this is an error
==============================
=== Experiment 2: Running `noisy_error` with `check=True`, and catching and tossing the Python exception
some info
this is an error
==============================
```

Not that there's anything wrong with the code you've written, it just looks odd to me.

My other questions are more important.

mweinelt:
1. No cleanup until reboot. We can revisit that in the future.
2. Added a print to stderr
3. Prints the stack trace now, but continues with the other AFI

PR #813, opened 2025-08-27
hydra: increase max_output_size (yet again)

vcunat:
https://github.com/NixOS/nixpkgs/pull/436667#issuecomment-3225514455

### DIFF
--- a/build/hydra.nix
+++ b/build/hydra.nix
@@ -112,7 +112,7 @@ in
 
     allow_import_from_derivation = false
 
-    max_output_size = 4121225472 # 3 << 30 + 900000000 = 3 GiB + 0.9 GB
+    max_output_size = 4294967295 # 4 GiB - 1 B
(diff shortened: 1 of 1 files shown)


vcunat:
Deployed already.  The point is to unblock `nixos-25.05`: https://hydra.nixos.org/build/306062350#tabs-buildsteps

vcunat:
Here and around: https://github.com/NixOS/nixos-channel-scripts/blob/master/mirror-nixos-branch.pl#L239

Mic92:
I think it would be more visible if the build done in a github action that can authenticate with S3. We would than no longer make this image a channel blocker because something would bump to the latest nixpkgs revision independently.

vcunat:
But right now it also works as a test aggregating many "basic packages", not just to create image.  I'm not sure about latest image being from a different commit than the latest channel.

PR #815, opened 2025-08-27
Update foundation emails

infinisil:
1. Offboard ex-board member @edolstra
2. Offboard @fricklerhandwerk from fundraising email
3. Redirect partnerships@nixos.org to foundation@nixos.org instead of just @refroni

1 and 2 are definitely fine, but I have to double check 3. Will mark as ready when done.


mweinelt:
Applied.

PR #816, opened 2025-08-28
builders aarch64: useTmpfs only for hopeful-rivest

vcunat:
We're getting out of RAM on goofy-hopcroft commonly: https://grafana.nixos.org/d/rYdddlPWk/node-exporter-full?orgId=1&var-job=node&var-node=goofy-hopcroft.builder.nixos.org:9100&from=now-30d&to=now&viewPanel=panel-78

While it does have double the RAM than hopeful-rivest, the 40 jobs can eat really lots of /tmp.
Some particular jobs even eat many gigabytes each.

As hopeful-rivest seems OK-ish, I'm leaving tmpfs in there for now: https://grafana.nixos.org/d/rYdddlPWk/node-exporter-full?orgId=1&var-job=node&var-node=hopeful-rivest.builder.nixos.org:9100&from=now-30d&to=now&viewPanel=panel-78

### DIFF
--- a/builders/profiles/hetzner-rx220.nix
+++ b/builders/profiles/hetzner-rx220.nix
@@ -1,10 +1,4 @@
 {
-  config,
-  lib,
-  ...
-}:
-
-{
   imports = [
     ../boot/efi-grub.nix
   ];
@@ -12,14 +6,6 @@
   disko.devices = import ../disk-layouts/efi-zfs-raid0.nix { };
   boot.supportedFilesystems.zfs = true;
   networking.hostId = "91312b0a";
-
-  # 96G for build roots, 160G for working memory
-  boot.tmp = {
-    useTmpfs = true;
-    # 128G tmpfs, 128G RAM for standard builders
-    # 128G tmpfs, 128G RAM for big parallel builders
-    tmpfsSize = if lib.elem "big-parallel" config.nix.settings.system-features then "128G" else "128G";
-  };
 
   boot.initrd.availableKernelModules = [
     "nvme"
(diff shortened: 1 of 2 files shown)


vcunat:
Maybe the mem PSI graphs are more precise for this purpose (after ctrl-clicking to show only the two reddish lines):
- https://grafana.nixos.org/d/rYdddlPWk/node-exporter-full?orgId=1&var-job=node&var-node=goofy-hopcroft.builder.nixos.org:9100&from=now-30d&to=now&viewPanel=panel-322
- https://grafana.nixos.org/d/rYdddlPWk/node-exporter-full?orgId=1&var-job=node&var-node=hopeful-rivest.builder.nixos.org:9100&from=now-30d&to=now&viewPanel=panel-322

builders/profiles/hetzner-rx220.nix
```diff
@@ -13,14 +7,6 @@
   boot.supportedFilesystems.zfs = true;
   networking.hostId = "91312b0a";
 
-  # 96G for build roots, 160G for working memory
-  boot.tmp = {
```
Mic92, line 17: It would be worth considering in this case having an additional /tmp zfs dataset with sync=disabled: https://github.com/NixOS/infra/blob/main/builders/disk-layouts/efi-zfs-raid0.nix

vcunat:
The current staging-next-25.05 is a bit stumped by this, because of a couple `gst-plugins-rs` builds that exhaust the RAM and retrying just repeats the incident.  So could we deploy this to `goofy-hopcroft` fast, ideally within a few hours?  _Honestly, I don't even know how this gets deployed, off the top of my head._

mweinelt:
You can just use nixos-rebuild with a remote target.

vcunat:
Oh, thanks, quite simple.  Deployed and rebooted that machine, building again now.

vcunat:
So far keeping this open, in case you want to deal with some nits.

PR #829, opened 2025-09-06
mailserver: Remove fpletz from SC

jtojnar:
He resigned https://discourse.nixos.org/t/my-departure-from-the-steering-committee/68903


mweinelt:
And the secret please.

PR #832, opened 2025-09-08
builders aarch64: do NOT useTmpfs, not even on hopeful-rivest

vcunat:
I was too optimistic.  After seeing repeated kills of Firefox builds during linking, I think this is more harm than good as it was. Perhaps some of the builds got more expensive over time.

The graph doesn't look good lately:
https://grafana.nixos.org/d/rYdddlPWk/node-exporter-full?orgId=1&var-job=node&var-node=hopeful-rivest.builder.nixos.org:9100&from=now-7d&to=now&viewPanel=panel-78

### DIFF
--- a/builders/instances/hopeful-rivest.nix
+++ b/builders/instances/hopeful-rivest.nix
@@ -9,12 +9,6 @@
     system-features = [ "big-parallel" ];
   };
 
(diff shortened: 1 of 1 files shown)


vcunat:
No reactions so far, but it's starting to get impractical that deployed and merged differs...

PR #839, opened 2025-09-12
fix: Listing of sub-directories of S3 buckets should work

LeSuisse:
This regression was introduced by 4030d7dc2585ec7ff52f2a80d9405c8a71fc93fa.

The mistake was made when "rebasing" the patch that was submitted to upstream to the version vendored in this repository. Upstream as a notion of base URL that is not present here.


mweinelt:
Applied. Thanks!

PR #840, opened 2025-09-13
Lock file: update nixpkgs

vcunat:
I wanted the builders to have a kernel with a bugfix that's affecting us: https://github.com/NixOS/nixpkgs/issues/438765


vcunat:
I already deployed this to our four linux builders and rebooted them.  (and restarted the older tailscale builds to cache them)

Issue #326, opened 2024-01-11
Migrate ngi0 infra to dedicated repository


mweinelt:
Hi @Ericson2314, @cleeyv, @lorenzleutgeb, @tomberek, @thufschmitt!

We talked today about the state of the infra with all infra members, and we agreed that it would be good to move ngi0 infra out of nixos-org-configurations. The reasons are that this would ultimately benefit you by unblocking your work on this infrastructure, as well as stop consuming our review bandwidth.

It would ultimately result in a better workflow for you guys, so what do you think?

The new home could be over at https://github.com/ngi-nix, if that is related, but we can also create a repo on the NixOS org.

WDYT?

lorenzleutgeb:
I was never bit by limited review bandwidth, and otherwise did not interact with the infra team much, so I am quite neutral, and trust your judgement that it would be better to move. However, are there any downsides of moving that we should be aware of?

I think that github.com/ngi-nix is related, and that it would be a good namespace. Mentioning owners of that org, for their comment: @fricklerhandwerk @mightyiam

mweinelt:
> However, are there any downsides of moving that we should be aware of?

None, that I am aware of.

Ericson2314:
I have no objections.

fricklerhandwerk:
Sounds fine to me. Would be reasonable to move that into `ngipkgs` once migrated, so it can live together with the code right there.

mightyiam:
Are we talking about moving it to a different runner computer?

mightyiam:
I wouldn't like my lack of context and possible catching up to delay whatever decision and follow up action. Just trying to learn.

zimbatm:
Ok, I created https://github.com/NixOS/ngi-infra as a placeholder to put the code in. We can decide where to move the repo later.

In this model, the underlying computer would stay the same, and the infra team would still hold access to that hardware and DNS records. At the same time, the NGI team can independently deploy configuration changes to their Hydra instance.

zimbatm:
Invited @fricklerhandwerk and @mightyiam to the https://github.com/orgs/NixOS/teams/ngi team. You should have permission to add missing team members.

fricklerhandwerk:
NGI0 infra config now lives in https://github.com/ngi-nix/ngi0-infra

PR #847, opened 2025-09-18
dns: Remove stale DNS records for an old NGI cache

Erethon:
My understanding is that in the past a Nix Cache used to be hosted in a S3 bucket that was populated by a Hydra instance running for the NGI team. This Hydra service is gone since at least [June 2024](https://github.com/ngi-nix/ngipkgs/commit/01a6dc8e4b2658ccba0c5842b6c795e9ee5cc68c).

The DNS records point to cloudfront and there [used to be terraform config](https://github.com/NixOS/infra/issues/326) for this cache that setup a S3 bucket (`ngi0-cache`). I don't have access to AWS, so I don't know if the bucket still exists, if it does, we should probably delete it as well as the cloudfront endpoint.

### DIFF
--- a/dns/nixos.org.js
+++ b/dns/nixos.org.js
@@ -129,9 +129,6 @@ D("nixos.org",
 	CNAME("freescout", "umbriel.nixos.org."),
 
 	// ngi
-	CNAME("cache.ngi0", "d2tu257wv37zz1.cloudfront.net."),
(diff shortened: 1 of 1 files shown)


jfly (approved):
> I don't have access to AWS, so I don't know if the bucket still exists, if it does, we should probably delete it as well as the cloudfront endpoint.

I also cannot check this. Hopefully someone else from the infra team can do this.

PR #852, opened 2025-09-21
Drop wireguard, source guard for Prometheus exporters

mweinelt:


modules/nftables.nix
```diff
@@ -0,0 +1,29 @@
...
+  };
+
+  networking.firewall = {
+    enable = true;
+
+    # be a good network citizen and allow some debugging interactions
+    rejectPackets = true;
+    allowPing = true;
+
+    # prevent firewall log spam from rotating the kernel rinbuffer
```
jfly, line 26: nnit: not your fault: `rinbuffer` -> `ringbuffer`

build/hydra.nix
```diff
@@ -15,10 +15,11 @@ in
     inputs.hydra.nixosModules.hydra
   ];
 
-  networking.firewall.allowedTCPPorts = [
-    9198 # queue-runnner metrics
-    9199 # hydra-notify metrics
-  ];
+  # queue-runner and hydra-notify metrics
+  networking.firewall.extraInputRules = ''
+    ip6 saddr $prometheus_inet6 tcp dport { 9190, 9199 }
```
jfly, line 20: Is this 9190 a typo? Should this be 9198 instead?

modules/prometheus/default.nix
```diff
@@ -35,9 +40,18 @@ in
     };
   };
 
+  networking.firewall.extraInputRules = ''
+    ip6 saddr $prometheus_inet6 tcp dport 9300
```
jfly, line 44: If we cannot interpolate the port number, I suggest we preserve the `# prometheus-nixos-exporter` comment from before.

jfly (approved):

mweinelt:
Applied.

PR #862, opened 2025-10-04
nixos.org.js : add an alias for an election-related service

7c6f434c:
We want to send clickable confirmation links with static target and list of clicked links in the server logs, so probably an alias to a personal VPS of an SC member should be the simplest


vcunat:
I fail to immediately see why the CI jobs are failing, but I'd rather end `michaelraskin.top` with a dot, so that it's clear that the name is not relative to the NixOS zone.

vcunat (dismissed):

7c6f434c:
Thanks. «I will just quickly edit this JS in the web editor, without making a local clone and also comparing to a usual zone file» was a mistake on my part, sorry

7c6f434c:
Thanks!

PR #863, opened 2025-10-07
Update NixOS logo

djacu:
Update the NixOS logo used treewide. Uses logo assets from brand.nixos.org. This URL should be stable for the foreseeable future.


jfly (approved):
Thanks!

vcunat (approved):

PR #885, opened 2025-11-04
Move test.wiki.nixos.org to more modern TLS configuration

mweinelt:
The new configuration comes with TLS1.2/1.3 and HTTP1.1/2/3 support.

The fastly webinterface only lists n.sni.global.fastly.net, but the dualstack hostname exists and I think it's worth giving it a shot.

image


vcunat:
Certainly sounds good in principle.

Mic92 (approved):

PR #916, opened 2025-12-14
Replace haumea with titan

mweinelt:
From:
- AMD Ryzen 7 3700X 8-Core Processor
- 64 GB DDR4 ECC memory
- 2x1TB NVME (zfs mirror, lz4)

To:
- AMD Ryzen 9 7950X3D 16-Core Processor
- 128 GB DDR5 ECC memory
- 2x2TB NVME (zfs mirror, zstd-3)

TODO:
- [x] rotate zrepl ssh key

### DIFF
--- a/build/haumea/network.nix
+++ b/build/haumea/network.nix
@@ -1,15 +1,6 @@
 {
   systemd.network = {
     enable = true;
-    netdevs = {
-      "20-vlan4000" = {
-        netdevConfig = {
-          Kind = "vlan";
-          Name = "vlan4000";
-        };
-        vlanConfig.Id = 4000;
-      };
-    };
     networks = {
       "30-enp35s0" = {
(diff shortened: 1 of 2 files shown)


mweinelt:
The new db host has been in production for 22 hours and it looks fine. Merging.

vcunat:
Nit: the migration broke the line of deduplication in zrepl backups.  So the backups consume roughly one full database of space in addition to the usual.  Permanently, or at least until we decide to wipe all <=2025 history.

While this makes a large relative jump in space usage of the backups, practically it won't matter, most likely, as we still have around 1.3 TiB free space on the backup host and I know of no plans to consume a lot more in there.

mweinelt:
That was kinda inevitable anyway due to the postgresql update.

PR #919, opened 2025-12-18
mailserver: Create email for the security tracker

Erethon:
This email address will be used only to send emails and for two uses cases in particular:
1) As a simple way for the developers of the software to get [notified about errors](https://github.com/NixOS/nix-security-tracker/issues/680) in the production instance.
2) In the future, send email notifications for potential vulnerabilities to maintainers (if they subscribe to them).

### DIFF
--- a/non-critical-infra/modules/mailserver/mailing-lists.nix
+++ b/non-critical-infra/modules/mailserver/mailing-lists.nix
@@ -169,6 +169,13 @@
       ];
     };
 
+    "noreply-securitytracker@nixos.org" = {
+      loginAccount = {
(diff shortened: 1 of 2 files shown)


non-critical-infra/modules/mailserver/mailing-lists.nix
```diff
@@ -169,6 +169,13 @@
       ];
     };
 
+    "securitytracker-noreply@nixos.org" = {
```
mweinelt:
```suggestion
    "noreply-securitytracker@nixos.org" = {
```

That would be my preference, so that people actually notice the noreply part. I do the same for DMARC and TLSRPT in SNM.

mweinelt (approved):

PR #921, opened 2025-12-20
channels: make nixpkgs-unstable primary

fricklerhandwerk:
For the [Nixpkgs security tracker](https://github.com/NixOS/nix-security-tracker/) UI we want to aggregate variants under their primary channels.
This can only be approximated heuristically, but `channels.nix` is authoritative and the closest to consistent we have.

So I wondered why `nixpkgs-unstable` is not primary. When this structure was established in #189, the field was simply ommited in https://github.com/NixOS/infra/pull/189#issuecomment-977137619.
Am I missing something from the semantics of "primary" that would forbid this change?

What makes a channel primary anyway? That the maximum number of tests is run?

### DIFF
--- a/channels.nix
+++ b/channels.nix
@@ -36,6 +36,7 @@ rec {
     };
     "nixpkgs-unstable" = {
       job = "nixpkgs/unstable/unstable";
+      variant = "primary";
       status = "rolling";
     };
 


fricklerhandwerk:
@NixOS/infra Do you think it makes sense to merge this, or is it too cosmetic to matter? I find it helpful for reasoning about channels if their variant isn't optional, and making `nixpkgs-unstable` "primary" would increase consistency.

channels.nix
mweinelt, line 14: The examples are not helping. 🤔

fricklerhandwerk: How about

```
When adding a new release, determine its variant:
<release>         => primary
<release>-small   => small
<release>-darwin  => darwin
<release>-aarch64 => aarch64
```

Is `aarch64` still a thing or can we remove that, too?

mweinelt: That leaves `release` underdefined. The aarch64-specific channel is gone. The darwin specific channels will be gone in 26.11. We'll eventually end up with

```
When adding a new release, determine its variant:

- primary:
  - all platforms, all packages (e.g. nixpkgs-unstable)
  - linux only, all packages (e.g. nixos-unstable)
- small: linux only, reduced package set (e.g. nixos-unstable-small)
- darwin: builds darwin only (e.g. nixpkgs-26.05-darwin)
```

And then we can make nixpkgs-unstable primary.

fricklerhandwerk: Hm, maybe it's a good opportunity to instead clarify the roles of the projects and jobsets? I know, that information is actually encoded in the Hydra database, in the specification of each jobset, but it would still help verbalising what they're for or what characterises them. In your examples that still stays somewhat obscure.

PR #986, opened 2026-03-24
builders: add support for systemd-nspawn containers

jfly:
This is part of https://github.com/NixOS/infra/issues/987. It gets
standalone nspawn containers working, but not tests that include both
nspawn containers and qemu vms.

We discussed this at our last team meeting, and folks didn't seem to
object to enabling these tests.

At the time, I thought it was a purely hypothetical conversation, but it
turns out that some of our jobsets in Hydra actually do [include these
new tests](https://github.com/NixOS/nixpkgs/pull/478109#issuecomment-4116714697).

### DIFF
--- a/builders/common/nspawn-test-containers.nix
+++ b/builders/common/nspawn-test-containers.nix
@@ -1,0 +1,17 @@
+{
+  # From <https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/development/running-nixos-tests.section.md#system-requirements-sec-running-nixos-tests-requirements>:
+  # > NixOS tests using systemd-nspawn containers require the Nix daemon to be
+  # > configured with the following settings:
+  nix.settings = {
+    auto-allocate-uids = true;
+    extra-system-features = [ "uid-range" ];
+    experimental-features = [
+      "auto-allocate-uids"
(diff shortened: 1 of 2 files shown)


builders/common/nspawn-test-containers.nix
```diff
@@ -0,0 +1,17 @@
+{
+  # From <https://github.com/NixOS/nixpkgs/blob/master/nixos/doc/manual/development/running-nixos-tests.section.md#system-requirements-sec-running-nixos-tests-requirements>:
+  # > NixOS tests using systemd-nspawn containers require the Nix daemon to be
+  # > configured with the following settings:
+  nix.settings = {
+    auto-allocate-uids = true;
+    extra-system-features = [ "uid-range" ];
+    experimental-features = [
+      "auto-allocate-uids"
+      "cgroups"
```
mweinelt, line 10: Do we require `use-cgroups = true`?

mweinelt: @jfly ping?

Mic92: @mweinelt No it's not required for the container feature. I used it without for years.

PR #995, opened 2026-04-02
macs: create darwin bootstrap closure

mweinelt:


jfly (approved):
I don't have experience with nix-darwin. Once we install a fancier profile than the bootstrap one, do we need to worry about the MDM custom command we wrote deploying the bootstrap configuration (thereby reverting the fancier profile)? 

LGTM!

macs/minimal.nix
```diff
@@ -92,4 +92,15 @@ in
     serviceConfig.StandardErrorPath = "/var/log/rosetta2-gc.log";
     serviceConfig.StandardOutPath = "/var/log/rosetta2-gc.log";
   };
+
+  # MacOS stores extensive logs in /var/db/uuidtext, which cause high disk usage
+  launchd.daemons.uuidtext-gc = {
+    script = ''
+      date
+      find /var/db/uuidtext -type f -mtime +7 -delete
```
mweinelt: could also be `log erase --all` in tahoe.

mweinelt:
> Currently I have the profile configured to continuously execute. So it would undo it yes. We could perhaps write the script to either use `bootstrap` or the static `hostname` if the hostname is not present in the flake we fallback to bootstrap?

That's not a blocker for this change, right?

This includes now two new workarounds that should make our life easier:
- regular log erase
- regular fseventsd killing

jfly:
I think this is good to merge as is. IIUC, it only becomes a problem if/when we have fancier profiles.

PR #1007, opened 2026-04-13
staging-hydra: use hydra branch with AWS_SHARED_CREDENTIALS_FILE fix

Mic92:
The queue-runner on the new lysator host could not load S3 credentials because the binary-cache crate ignored AWS_SHARED_CREDENTIALS_FILE and fell back to the (unreachable) EC2 IMDS endpoint, stalling all narinfo lookups. Point hydra-staging at Mic92/hydra:aws-shared-credentials-file until the upstream PR lands.

### DIFF
--- a/build/titan/zrepl.nix
+++ b/build/titan/zrepl.nix
@@ -107,15 +107,15 @@ in
               format = "human";
             }
           ];
-        };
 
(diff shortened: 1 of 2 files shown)


jfly (approved):
Thanks!

PR #1008, opened 2026-04-13
nix: switch back to release

Mic92:


flake.nix
```diff
@@ -16,7 +16,7 @@
     # errors when curl reuses connections that S3 has already dropped.
     # Remove nix override once hydra's nix input includes the upstream revert.
```
jfly: Is this comment now rotten?

jfly (approved):

flake.nix
Mic92: Good catch.

PR #1012, opened 2026-04-13
Alert if zrepl replication hasn't succeeded recently

jfly:
zrepl is currently configured to take snapshots every 30 minutes, so if we go a full day without a successful replication, I'd say something is pretty wrong.


Mic92:
deployed.

PR #1015, opened 2026-04-20
build: allow releases.nixos.org in restricted-eval

Mic92:
Nix's hydra.nix pins nixpkgs via a tarball on releases.nixos.org. Without this URI prefix in allowed-uris, restricted-mode evaluation rejects the fetch and all jobs in the jobset fail.


Mic92:
Background is that we want to have faster nix security releases: https://github.com/NixOS/nix/pull/15640

Eveeifyeve (approved):

PR #1020, opened 2026-05-01
haumea: allow bmg to login

vcunat:


vcunat:
I just switched the identifier to`brianmcgillion`, as that's used on github.com and in nixpkgs.

Overall I see no motivation to delay this.  Haumea isn't important anymore.  And not addressing this in any way will cause other deploys to prevent Brian from accessing the machine &ndash; again &ndash; which would be pretty annoying.

PR #1025, opened 2026-05-04
haumea: use the correct identifier for BMG

vcunat:
Otherwise it could be really confusing.


vcunat:
I'm sorry about the naming confusion in PR https://github.com/NixOS/infra/pull/1020

PR #1028, opened 2026-05-08
nft-prefix-import: follow the usual branch from now on

vcunat:
The fix was merged long ago; risk of forgetting to unpin. https://github.com/mweinelt/nft-prefix-import/pull/2

This is basically a partial revert of commit 7d400ce2e.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -349,16 +349,15 @@
         ]
       },
       "locked": {
-        "lastModified": 1776722121,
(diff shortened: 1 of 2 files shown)


vcunat:
> error: cannot update unlocked flake input 'nft-prefix-import' in pure mode

Uh, am I doing this in a wrong way?  🤔

jfly:
@vcunat, `nix flake update nft-prefix-import` does the right thing for me locally when I check out this branch.

vcunat:
Perhaps the CI assumes that I should also update the lock in such cases.

jfly (approved):

PR #1030, opened 2026-05-09
hydra: increase upload attempts

mweinelt:
Workaround missing outputs causing an incompatible rebuild that leads to SIGKILL on darwin.

Ref: https://github.com/NixOS/nix/pull/15638#issuecomment-4413076030

cc @emilazy

### DIFF
--- a/build/hydra.nix
+++ b/build/hydra.nix
@@ -21,17 +21,23 @@ in
     ip saddr $prometheus_inet4 tcp dport { 9198, 9199 } accept
   '';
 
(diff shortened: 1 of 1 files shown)


emilazy (approved):
🤞

A more robust solution would be good, but since the new queue runner will presumably handle uploads completely differently this seems good to try for now.

I do worry about whether the kinds of network conditions that cause 5 uploads in a row to fail will also cause 32 of them to fail, but I guess we can iterate if we see it continue to crop up.

mweinelt:
Applied.

PR #1031, opened 2026-05-11
dns: update norwegian-blue.mac.nixos.org

vcunat:


vcunat:
The old IPv6 stopped working.  After discussion with Flying Circus (which hosts this machine) we're switching to this one which is expected to be stable.

PR #1033, opened 2026-05-14
hydra-mirror: migrate from classic PAT

mweinelt:
This shortly broke while we tried disabling classic PATs on the NixOS org on GitHub. With PATs being limited in lifetime we are going for an SSH key instead, which also limits us to Git operations and no API acccess.

Tested and the update services do at least start up.

Ref: https://github.com/NixOS/org/issues/247


modules/hydra-mirror.nix
```diff
@@ -39,6 +39,9 @@ let
         User = "hydra-mirror";
         # Allow the unit to use 80% of the system's RAM and 100% of the system's swap
         MemoryHigh = "80%";
+        LoadCredential = [
```
jfly, line 42: I don't understand why we didn't need this before. Does this job run as root? Or is there some filesystem ownership on this file we could remove now?

jfly (approved):

modules/hydra-mirror.nix
jfly: Oh duh, I see you removed the owner below.

mweinelt: This makes it available in the scope of the systemd unit only, not in the scope of the user.

PR #1039, opened 2026-05-17
macs: reuse shared code with ofborg macs

mweinelt:
This deduplicates code, simplifies maintenance and applies important workarounds to make Macs healthy builders with high throughput.


keys.nix
```diff
@@ -42,12 +42,22 @@ rec {
...
         "ecdsa-sha2-nistp256 AAAAE2VjZHNhLXNoYTItbmlzdHAyNTYAAAAIbmlzdHAyNTYAAABBBKHHl5kgMDNQA/zqK+AzT4SO09rfAp+y/EeUC+Ow5XqyNid5lm6sgLGM+AqZDx0jOrMKWhd5lhzGDdtsSf0Y8g4= brian@saturn"
       ];
+
+      janne = [
+        "ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAIM35Bq87SBWrEcoDqrZFOXyAmV/PJrSSu3hl3TdVvo4C janne"
+      ];
+
+      conni2461 = [
+        "ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAIPK/3rYhlIzoPCsPK38PMdK1ivqPaJgUqWwRtmxdKZrO ✏️"
+      ];
```
mweinelt, line 52: @Conni2461 Can you confirm this is your key?

mweinelt: Got confirmation over Matrix.

mweinelt:
```
error: config contains the wrong home directory for hydra-queue-builder, aborting activation
nix-darwin does not support changing the home directory of existing users.

Please set:

    users.users.hydra-queue-builder.home = "/private/var/lib/hydra-queue-builder";

or remove it from your configuration.
```

Forcing might be required here.

Mic92:
@Enzime probably knows what to do.

mweinelt:
We are already forcing the homedir for node-exporter, so I'm confident the same approach will work.

mweinelt:
Well no. Setting that does nothing. Same assertion.

PR #1048, opened 2026-05-22
ofborg: add support for nspawn containers

zowoq:
cc @jfly @mweinelt 

I guess these should be enabled on ofborg as well now that they are enabled on h.n.o?


jfly (approved):
Sounds reasonable to me!

mweinelt:
Will deploy in an hour if noone beats me to it.

mweinelt:
Applied.

PR #1054, opened 2026-05-25
NixOS 26.05

mweinelt:
Updating everything to NixOS 26.05. We only depend on -small anyhow, so we can start working on the upgade.

Everything builds.

### DIFF
--- a/builders/common/nix.nix
+++ b/builders/common/nix.nix
@@ -25,7 +25,6 @@
       builders-use-substitutes = true;
       extra-experimental-features = [
         "nix-command"
-        "no-url-literals"
         "flakes"
       ];
       system-features = [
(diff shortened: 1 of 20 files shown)


builders/common/nix.nix
```diff
@@ -25,7 +25,6 @@
       builders-use-substitutes = true;
       extra-experimental-features = [
         "nix-command"
-        "no-url-literals"
```
xokdvium, line 28: Shouldn't this enable the option it has been replaced by? Though I'm not sure how useful it is, since the builders don't do eval.

mweinelt: Yeah, not super useful for the builders at all. I would expect nixpkgs-ci to bark at people for the most important jobsets.

xokdvium: Makes sense.

mweinelt:
Applied and rebooted.

PR #1061, opened 2026-05-29
mail: roll new dkim keys

mweinelt:
The existing RSA key is 1024 bit long and needs to be retired.

We generate the new keys in place, managing them is imo pretty pointless, as we won't be able to sign new mail that will get validated anyhow. At which point we'll publish new selectors.


mweinelt:
The old API key was not working any more. The new one for some reason does not include ofborg.org. I've reached out to the account owners to check this out.

PR #1062, opened 2026-05-30
Point DNS for docuseal.nixos.org to caliban

jfly:
I'd like to merge this first so I can iterate on umbriel with a valid certificate.

This is part of https://github.com/NixOS/infra/issues/972


mweinelt:
Why would you keep stacking things onto the mail host? Caliban is the better choice for apps.

jfly:
@mweinelt, I didn't know. Updated.

PR #1066, opened 2026-06-02
apply Nix patches to (try to) fix S3 upload abortions

vcunat:


vcunat:
Deployed to `mimas`.

vcunat:
Let's merge even if it's a temporary state.  Otherwise the other deploy actions might interfere, etc.

vcunat:
Deployed everywhere (except haumea being problematic generally)

PR #1078, opened 2026-06-15
mailserver: fix "dovecot.service" naming

vcunat:
We sometimes got errors, e.g.
> sops-install-secrets: cannot request units to restart: error running 'systemctl --no-block try-restart dovecot2.service': exit status 5


vcunat:
Deployed to `umbriel`.

PR #1085, opened 2026-06-18
hydra: raise maxUnsupportedTimeInS to 24h

Mic92:
Builders register over gRPC, so a reconnect or network blip briefly drops their system from the queue runner. With the 120s default, abort_unsupported then fails buildable steps as "Unsupported system type" (NixOS/hydra#1805).


vcunat (approved):
I see no risks from increasing this, only potential benefits.

vcunat:
Uh, (seeing the CI in this PR) the hash mismatches might even complicate deploying fixes to the infra...

PR #1099, opened 2026-06-22
hydra: re-introduce a connection pool for read-only valid path queries

Mic92:
https://github.com/NixOS/hydra/commit/2f8214acda4aea99876ed5b85b7994ff7e6f7b92

Should be merged after and testing in staging building NixOS unstable small

In our ingestion code we spawn now a lot of daemon connections.
Initially we had a daemon pool but that one had a bug where wedged
connections were returned.
This patch is way more conservative and reduces the live time of a pool
and limit it to basically sqlite is_valid_path queries.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -28,7 +28,7 @@
     };
 
     hydra = {
-      url = "github:NixOS/hydra/hydra.nixos.org";
+      url = "github:NixOS/hydra/staging-hydra.nixos.org";
       inputs.nixpkgs.follows = "nixpkgs";
       inputs.nix.follows = "nix";
       inputs.treefmt-nix.follows = "treefmt-nix";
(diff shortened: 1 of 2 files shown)


mweinelt:
This now runs in prod after staging looked better.

flake.nix
```diff
@@ -28,7 +28,7 @@
     };
 
     hydra = {
-      url = "github:NixOS/hydra/hydra.nixos.org";
+      url = "github:NixOS/hydra/staging-hydra.nixos.org";
```
mweinelt, line 31: I think the hydra.nixos.org branch is caught up, can you revert and relock?

vcunat:
These changes got pulled automatically during https://github.com/NixOS/infra/pull/1109

PR #1103, opened 2026-06-25
nixos-channel-scripts: add build product disambiguation by regex pattern

mweinelt:
When a build has multiple build products of the same type the script
would previously bail out. Now we allow passing a regular expression to
filter down the list of build products to only matching ones.

This is a requirement before we introduce a zstd compressed nixexprs tarball.

### DIFF
--- a/pkgs/nixos-channel-scripts/mirror-nixos-branch.pl
+++ b/pkgs/nixos-channel-scripts/mirror-nixos-branch.pl
@@ -225,7 +225,7 @@ if ($bucketReleases && $bucketReleases->head_key("$releasePrefix")) {
     }
 
     if ($channelName =~ /nixos/) {
-        downloadFile("nixos.channel", "nixexprs.tar.xz");
+        downloadFile("nixos.channel", "nixexprs.tar.xz", "source-dist", '\.tar\.xz$');
         downloadFile("nixpkgs.tarball", "packages.json.br", "json-br");
         downloadFile("nixos.options", "options.json.br", "json-br");
 
@@ -265,7 +265,7 @@ if ($bucketReleases && $bucketReleases->head_key("$releasePrefix")) {
         }
 
     } else {
-        downloadFile("tarball", "nixexprs.tar.xz", "source-dist");
+        downloadFile("tarball", "nixexprs.tar.xz", "source-dist", '\.tar\.xz$');
         downloadFile("tarball", "packages.json.br", "json-br");
     }
 
--- a/pkgs/nixos-channel-scripts/mirror-nixos-branch.pl
+++ b/pkgs/nixos-channel-scripts/mirror-nixos-branch.pl
@@ -156,24 +156,34 @@ if ($bucketReleases && $bucketReleases->head_key("$releasePrefix")) {
     }
 
     sub downloadFile {
-        my ($jobName, $dstName, $productType) = @_;
+        my ($jobName, $dstName, $productType, $productPattern) = @_;
 
         my $buildInfo = decode_json(fetch("$evalUrl/job/$jobName", 'application/json'));
 
         my $products = ();
         # Key the products by subtype.
         foreach my $key (keys $buildInfo->{buildproducts}->%*) {
-            my $subType = $buildInfo->{buildproducts}->{$key}->{subtype};
+            my $product = $buildInfo->{buildproducts}->{$key};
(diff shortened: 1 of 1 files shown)


pkgs/nixos-channel-scripts/mirror-nixos-branch.pl
```diff
@@ -156,24 +156,34 @@ sub fetch {
...
+            my $product = $buildInfo->{buildproducts}-{key};
+            my $subType = $product->{subtype};
+
+            next if defined $productPattern
+                 && $product->{path} !~ /$productPattern/;
+
             if ($products->{$subType}) {
-                die "Job $jobName has multiple products of the same subtype $subType.\nThis is a bad assumption from this script";
+                if (defined $productPattern) {
+                    die "Job $jobname has multiple products of subtype $subType that match $productPattern.\nRefine the product regex pattern further to disambiguate.";
```
Erethon: s/jobname/jobName/?

pkgs/nixos-channel-scripts/mirror-nixos-branch.pl
```diff
@@ -156,24 +156,34 @@ sub fetch {
...
+                } else {
+                    die "Job $jobName has multiple products of the same subtype $subType.\nPass a product regex pattern to disambiguate.";
+                }
             }
-            $products->{$subType} = $buildInfo->{buildproducts}->{$key};
+
+            $products->{$subtype} = $product;
```
Erethon: s/subtype/subType/

pkgs/nixos-channel-scripts/mirror-nixos-branch.pl
```diff
@@ -156,24 +156,34 @@ sub fetch {
...
-        my ($jobName, $dstName, $productType) = @_;
+        my ($jobName, $dstName, $productType, $productPattern) = @_;
 
         my $buildInfo = decode_json(fetch("$evalUrl/job/$jobName", 'application/json'));
 
         my $products = ();
         # Key the products by subtype.
         foreach my $key (keys $buildInfo->{buildproducts}->%*) {
-            my $subType = $buildInfo->{buildproducts}->{$key}->{subtype};
+            my $product = $buildInfo->{buildproducts}-{key};
```
Erethon: s/-{key/->{$key/

Erethon (commented):
I didn't run this, reviewed by the diff only.

xokdvium:
> before we introduce a zstd compressed nixexprs tarball.

Hm, what's the plan there? I don't think nix supports anything other than bzip2/xz for channel tarballs in the first place?

mweinelt:
You are saying that libarchive support only exists for NARs?

https://github.com/NixOS/nixpkgs/pull/535272

xokdvium:
I'm talking about this:

https://github.com/NixOS/nix/blob/master/src/nix/nix-channel/nix-channel.cc#L158C17-L164C18

I guess you could just consume it via other means (flake inputs and whatnot), but this won't improve the channel based tooling. I guess I might just be misunderstanding the intention.

https://github.com/NixOS/nix/blob/master/src/libfetchers/tarball.cc#L243C1-L251C6

Curiously, builtin unpack-channel supports all types of compression.

https://github.com/NixOS/nix/blob/master/src/libstore/builtins/unpack-channel.cc#L25C5-L25C29

mweinelt:
Fair, I did not have channels in mind as a consumer, after browsing the GitHub-brained Flake registry.

I don't think anything prevents us from adding zst support and start discontinuing tar.xz tarballs with the EOL of NixOS 27.05 (2027-12-31) or 27.11 (2028-06-30).

xokdvium:
> and start discontinuing tar.xz tarballs with the EOL of NixOS 27.05

Hm, are you saying that we should add zstd support to `nix-channel` tooling now?

mweinelt:
Looks like one more try/catch layer with a preference for zstd once we have it in place. Is that an outrageous ask?

xokdvium:
No, it just wasn't clear. A bit of a compact hazard, but it should be good. I guess the first step is to benchmark and see just how much of a benefit we'd get (last I recall decompression wasn't actually that big of an issue with nix, more tarball unpacking + libgit2 writes).

mweinelt:
Tested and nixos-26.05-small advanced.

PR #1104, opened 2026-06-26
Backport nix temproot gc fix and queue builder temproot upload fix

Mic92:


booxter:
Do mac builders need the same backport for Nix? Looks like they have a separate config under macs/.

mweinelt:
Applied.

PR #1105, opened 2026-06-26
ofborg: return to release branch

mweinelt:


mweinelt:
They're the same commit.

vcunat:
> They're the same commit.

The `rev:` line in `flake.lock` changed, so they're not?  Do I misunderstand you?

mweinelt:
They have the same commit on top. The commit id changed.

vcunat:
Ah right, I see.  The two commits do have the same file-set, except for nixpkgs version in their `flake.lock`.

PR #1108, opened 2026-06-27
builders, macs: migrate to fast-nix-gc

mweinelt:
https://github.com/Mic92/fast-nix-gc

### DIFF
--- a/macs/common/nix.nix
+++ b/macs/common/nix.nix
@@ -1,10 +1,15 @@
 {
   config,
+  inputs,
   pkgs,
   ...
 }:
 
 {
+  imports = [
+    inputs.fast-nix-gc.darwinModules.default
+  ];
+
   environment.systemPackages = [
     config.nix.package
   ];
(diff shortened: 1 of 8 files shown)


macs/common/nix.nix
```diff
@@ -23,18 +28,16 @@
...
-        }
-      ];
-      # ensure up to 100G free space every half hour
-      options = "--max-freed $(df -k /nix/store | awk 'NR==2 {available=$4; required=100*1024*1024; to_free=required-available; printf \"%.0d\", to_free*1024}')";
-    };
   };
+
+  services.fast-nix-gc = {
+    enable = true;
+    automatic = true;
```
Mic92, line 35: Afaik with queue builder, all our nix-daemon connections on builders are short lived. But if we see strange WAL behaviour (big files that don't shrink), we might need this: https://github.com/Mic92/fast-nix-gc#when-to-use---no-vacuum--novacuum

mweinelt:
Deployed to all builders.

PR #1110, opened 2026-06-29
hydra: update, improving GC safety

vcunat:
This only ff-pulls a single commit:
  [469074f2](https://github.com/NixOS/hydra/commit/469074f2)  queue-runner: create GC-safe perm roots for build outputs

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -436,11 +436,11 @@
         ]
       },
       "locked": {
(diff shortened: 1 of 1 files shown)


vcunat:
I confirmed via direct communication that this is suitable for deployment but won't really help our use case.

PR #1111, opened 2026-06-29
dns: add surveys.nixos.org

djacu:
In preparation for the new surveys page.

### DIFF
--- a/dns/nixos.org.js
+++ b/dns/nixos.org.js
@@ -217,6 +217,7 @@ D("nixos.org",
 	CNAME("planet", "nixos-planet.netlify.app."),
 	CNAME("search", "nixos-search.netlify.app."),
 	CNAME("status", "nixos-status.netlify.app."),
+	CNAME("surveys", "nixos-surveys.netlify.app."),
 	CNAME("weekly", "nixos-weekly.netlify.com."),
 	CNAME("www", "nixos-homepage.netlify.app."),
 );


mweinelt:
Is there any implication for the limesurvey instance in here that I am missing?

djacu:
I completely forgot we host lime survey at survey.nixos.org. Putting the survey results at surveys.nixos.org would be confusing. Any ideas for a better subdomain?

mweinelt:
I suppose a category on Discourse is not what you are looking for?

jfly:
I wonder if we could use the survey.n.o domain for both limesurvey and whatever results website you're setting up. To create space for that, I think we could "move" limesurvey into a subpath (/survey? /lime?, IDK). That would make things more discoverable, IMO. Here's the relevant limesurvey docs: https://www.limesurvey.org/manual/Optional_settings#URL_settings, seems like you need to read up on the Yii framework to figure out how to do it.

djacu:
> I suppose a category on Discourse is not what you are looking for?

This year I used echarts and astro to create an webpage for the survey results. I am hoping that more teams can use the same pipeline in the future to process their survey results and post them on the shared survey site.

image

djacu:
Hey @mweinelt @jfly 
I dug into the limesurvey docs and it looks like implementing it would get messy.

@thilobillerbeck did some work to pull the surveys repo into the nixos-homepage repo and deploy it directly to nixos.org/surveys.

Appreciate all the ideas and help. I don't think there is any reason to keep this PR open so I'll close it for now.

PR #1112, opened 2026-06-29
builders: make big-parallel not mandatory on sleepy-brown

mweinelt:
This is a test to see if we can keep it doing something and still get the big-parallel builds through.

### DIFF
--- a/builders/instances/sleepy-brown.nix
+++ b/builders/instances/sleepy-brown.nix
@@ -8,7 +8,7 @@
     max-jobs = 4;
   };
 
-  services.hydra-queue-builder-dev.mandatoryFeatures = [ "big-parallel" ];
+  services.hydra-queue-builder-dev.supportedFeatures = [ "big-parallel" ];
 
   networking = {
     hostName = "sleepy-brown";


vcunat:
We could try how this goes, but:
- for that it would be great if we had some visibility into lengths of big-parallel sub-queues again
- so far the machine has been fully occupied most of the time even without this ([here](https://grafana.nixos.org/d/MJw9PcAiz/hydra-jobs?from=now-7d&to=now&viewPanel=panel-14) + click-select that particular machines)
- about the same goes for the aarch64 one (`hopeful-rivest`)

vcunat:
As `sleepy-brown` was almost fully idle during the past several hours, I now deployed it there (rebased atop the current `main`).

mweinelt:
Yeah, just remember that it resets every day to what is on main.

The four jobs didn't really put any load on sleepy-brown either, when I tried it recently.

I feel like we may need a more complicated build slot specification.

vcunat:
I'd say let's merge this until we suspect issues.  It's better to do some work than be idle.

mweinelt:
Feel free to approve and merge.

vcunat (approved):

vcunat:
Ah, I forgot that it needs approval.

mweinelt:
> Ah, I forgot that it needs approval.

Not trying to gaslight you or anyone, so I'll tell you that I re-enabled that requirement, after I saw our review wasn't required for DNS changes any more, while

```
❯ head -n2 .github/CODEOWNERS
# Every directory containing configurations impacting the core infra needs a
# review from a member of core infra.
```

Not sure who turned that off, but we never talked about turning it off.

PR #1124, opened 2026-07-08
nixos-channel-scripts: fix disambiguation logic

mweinelt:
Previously it would error out if any subtype had multiple files that were not disambiguated instead of just the subtype we were interested in.


mweinelt:
Tested and works as intended.

PR #1125, opened 2026-07-08
CODEOWNERS: add nixos-channel-scripts

mweinelt:


vcunat (approved):

vcunat:
One space gets added on every line, so that makes the diff a little harder to read but still fine.

emilylange:
> One space gets added on every line, so that makes the diff a little harder to read but still fine.

Note that you can "Hide Whitespaces" in "Open diff view settings": https://github.com/NixOS/infra/pull/1125/changes?w=1

vcunat:
Ah, nice.  Though I used and prefer local tools (outside the browser).

PR #1163, opened 2026-08-01
mailserver: order postfix-setup after sops-install-secrets

### LINKED ISSUE #1047: Race condition between `sops-install-secrets` and `postfix-setup-start`
Seen on umbriel. Note how `sops-install-secrets` starts, `postmap` crashes, and then `sops-install-secrets` finishes.

```
May 19 06:29:49 umbriel sops-install-secrets[661]: sops-install-secrets: Imported /etc/ssh/ssh_host_ed25519_key as age key with fingerprint age15vcp7875xwtf64j4yshyld0a3hpgzv6n2kxky493s3q0swr9hdaqxugpv6
May 19 06:29:49 umbriel postfix-setup-start[925]: postmap: fatal: open /var/lib/postfix/conf/virtual-mailing-lists: No such file or directory
May 19 06:29:49 umbriel postfix/postmap[925]: fatal: open /var/lib/postfix/conf/virtual-mailing-lists: No such file or directory
...
May 19 06:29:49 umbriel systemd[1]: Finished sops-install-secrets.service.
```

efadeev:
Closes #1047

### DIFF
--- a/non-critical-infra/modules/mailserver/mailing-lists-options.nix
+++ b/non-critical-infra/modules/mailserver/mailing-lists-options.nix
@@ -136,6 +136,10 @@ in
       restartUnits = [ "postfix-setup.service" ];
     };
 
+    systemd.services.postfix-setup = {
+      after = [ "sops-install-secrets.service" ];
+    };
+
     services.postfix.mapFiles.virtual-mailing-lists =
       config.sops.templates."postfix-virtual-mailing-lists".path;
 


jfly (commented):
Thanks, @efadeev. This looks plausible.

WDYT about [this other idea](https://github.com/NixOS/infra/issues/1047#issuecomment-4801780997)?

> So the fix might be to just add a `wantedBy = [ "multi-user.target" ];` to `postfix-setup.service`

If it works, I find it more appealing because it would be something we could upstream to nixpkgs, thus relieving us of having to maintain this code here, and also save others from running into the same issue.

efadeev:
Thanks for the review, @jfly!

I dug into this and I don't think `wantedBy = [ "multi-user.target" ]` on `postfix-setup.service` would actually fix the race. I think the missing ordering is one level down, in sops-nix.

modules/sops/default.nix#L467
```
 wantedBy = [ "sysinit.target" ];
 after = [ "local-fs.target" "systemd-sysusers.service" "userborn.service" ];
 requiredBy = [ "sysinit-reactivation.target" ];
 before = [ "sysinit-reactivation.target" ];   # note: no `before` on
 sysinit.target
 unitConfig.DefaultDependencies = "no";
```

Normally a target implicitly gains `After= on` everything it `Wants=`, which is what you'd expect to order this before `sysinit.target`. But systemd skips that when either side opts out of default dependencies. 

Since `sops-install-secrets.service` sets `DefaultDependencies=no`, `sysinit.target` never gets `After=sops-install-secrets.service`. So the unit is pulled into the boot transaction but has no ordering at all relative to `sysinit.target` / `basic.target` and everything downstream of them. That's the race.

Compare with `userborn.service` in nixpkgs (nixos/modules/services/system/userborn.nix#L138), which has the same shape (`DefaultDependencies = false`, `wantedBy = [ "sysinit.target" ]`) but spells the ordering out explicitly:

```
before = [ "systemd-tmpfiles-setup-dev.service" "sysinit.target"
"shutdown.target" "sysinit-reactivation.target" ];
conflicts = [ "shutdown.target" ];
```

That's the standard pattern for early-boot units, and it's exactly what sops-nix is missing.

Why `wantedBy = [ "multi-user.target" ]` doesn't close it: `postfix-setup.service` already gets implicit `After=sysinit.target` basic.target from `DefaultDependencies=yes`. Adding a `wantedBy` only changes what pulls it in, not its ordering relative to `sops-install-secrets.service:` no new ordering edge is created between the two units.

If anything it could widen the window: today postfix-setup is pulled in via `Requires=` from `postfix.service`, which is also `After=network.target`. Making it `wantedBy = multi-user.target` lets it start as soon as basic.target is up, i.e. earlier.

I think the upstreamable fix is in sops-nix rather than nixpkgs: add `before = ["sysinit.target" "shutdown.target" ]` (plus `conflicts = [ "shutdown.target" ]`) to `sops-install-secrets.service`, mirroring userborn. That fixes it for every secret consumer at once like for dovecot, postsrsd, mailman etc., not just postfix.

jfly:
I am sorry, but I have not read your whole message carefully. Responding to the most important part:

> I think the upstreamable fix is in sops-nix rather than nixpkgs

Let's pursue that instead of a workaround over here.

efadeev:
> Let's pursue that instead of a workaround over here.

OK, I'll try to do it later

PR #1169, opened 2026-08-07
mailserver: drop nixpkgs-core mailing list

emilazy:
The Nixpkgs core team [has disbanded](https://github.com/NixOS/org/pull/277).

It looks like we can’t remove FreeScout access ourselves, so that will need to be handled by an admin.


vcunat:
Looks OK, deployed to umbriel.

PR #1178, opened 2026-08-19
channels: improve flake-registry caching resilience

Mic92:
GitHub only sends max-age=300 for the flake registry, so the cached copy could be evicted and stale_if_error had nothing to serve when GitHub errored or rate-limited.

Cache for a day, serve the stale copy on any >=400 from GitHub, and revalidate in the background so clients never block on GitHub.

### DIFF
--- a/terraform/channels.tf
+++ b/terraform/channels.tf
@@ -222,7 +222,15 @@ resource "fastly_service_vcl" "channels" {
         set beresp.cacheable = true;
       }
       if (req.url ~ "/flake-registry.json") {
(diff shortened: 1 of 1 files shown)


Mic92:
I haven't actually validated the syntax/code yet but let's discuss these defaults on a conceptional level.

vcunat:
I do like stale-serving this on GitHub errors.

At a quick glance, this should mainly help the flakes which map to channels.nixos.org as the others currently in there go to github.com anyway.

PR #1180, opened 2026-08-21
mailserver: add nix.oss-fuzz mailing list

xokdvium:
To accomodate https://github.com/google/oss-fuzz/pull/16033.

### DIFF
--- a/non-critical-infra/modules/mailserver/mailing-lists.nix
+++ b/non-critical-infra/modules/mailserver/mailing-lists.nix
@@ -232,5 +232,14 @@
         ../../secrets/sigmasquadron-xsa-email-address.umbriel # https://github.com/SigmaSquadron
       ];
     };
+
+    # Primary point of contact for https://github.com/google/oss-fuzz reports
+    # for https://github.com/NixOS/nix. Subset of https://nixos.org/community/teams/nix.
+    "nix.oss-fuzz@nixos.org" = {
+      forwardTo = [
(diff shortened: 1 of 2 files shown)


xokdvium:
@NixOS/nix-team feel free to add (request to add yourself here, I can give up push right for my fork if needed).

xokdvium:
Hm sucks, apparently they require a google account huh https://google.github.io/oss-fuzz/faq/#why-do-you-require-a-google-account-for-authentication.

jfly:
> Hm sucks, apparently they require a google account

Darn. So the mailing list you've added here won't do us any good?

xokdvium:
> Darn. So the mailing list you've added here won't do us any good?

I'm not sure. Maybe we can set up a shared account with secrets in the vault (for the issue tracker) and have it set up to forward to this list?

JamieMagee:
An OSS Fuzz maintainer suggested using the `primary_contact` in addition to `auto_ccs` list instead of having to figure out how to get a mailing list to work[^1]. What do you think?

[^1]: https://github.com/google/oss-fuzz/pull/16033#discussion_r3847008623

vcunat:
> apparently they require a google account

Addressed here via 2e43b056e66733?

JamieMagee:
`google-site-verification` TXT records verify that you own a domain[^1]. It's unrelated to a Google account.

[^1]: https://support.google.com/webmasters/answer/9008080?hl=en

vcunat:
Ah right, that was for Google Workspace.  I'm sorry for the noise; let me hide these comments.

PR #1183, opened 2026-08-25
macs: deploy as root@

vcunat:
The other names (often) don't work for me.


vcunat:
Maybe this script is an outdated way of deploying?  Anyway, it seems to work OK after this PR.

PR #1185, opened 2026-08-26
Introduce separate flake inputs for staging hydra

Ericson2314:
We want to see if Hydra with Nix 2.35 works again without testing in prod for once!

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -113,6 +113,21 @@
         "type": "github"
       }
     },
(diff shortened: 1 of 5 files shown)


Mic92 (approved):
not yet deployed but this come later (staging hydra only)

vcunat:
Deployed later; discussed in https://matrix.to/#/#hydra:nixos.org

PR #1194, opened 2026-09-02
draupnir: use nodejs_26 to avoid crashes

vcunat:


vcunat:
Deployed.  The logs look OK to me so far.

vcunat:
It's basically taken from PR https://github.com/NixOS/nixpkgs/pull/558283

PR #1196, opened 2026-09-05
staging-hydra: follow hydra `master`

Ericson2314:
This input has been pointing at a succession of PR branches, and all those PRs checked out and now are merged. So with the bump, staging hydra should accordingly have these configuration changes:

- eval error mail notifications set up to a dummy log file, not a real MTA. (build error notifications are still disabled.)

- new websocket server

- `max_concurrent_evals` moved to the evaluator's own TOML file, so it is not left behind in `hydra.conf` silently becoming the default.

This is already deployed (I can do staging hydra though not the builders). I am just sending this PR to reflect what was done.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -491,16 +491,16 @@
         ]
       },
       "locked": {
-        "lastModified": 1787803693,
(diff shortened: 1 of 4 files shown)


mweinelt:
Can we split that up into one commit per concern?

PR #1197, opened 2026-09-06
Use channels.nixos.org tarballs in flake inputs

GraysonTinker:
I assume you all don't need to get told the benefits of this :)


vcunat:
A nice suggestion 🙂

Deployed via `colmena apply`.  I notice no issues.

PR #1209, opened 2026-09-13
wiki: stop 429ing anonymous Special:UserLogin/CreateAccount at the edge

Lassulus:
Since the Fastly cutover, nobody can log in to or register on wiki.nixos.org.

The `recv` snippet returns `error 429` for any request without a session cookie to `/w/index.php?title=Special:UserLogin|CreateAccount` when `returnto=` is present. Every "Log in" / "Create account" link in the UI carries `returnto=`, and a logged-out user by definition has no session cookie, so the rule blocks the whole login flow rather than scrapers:

```
$ curl -sI 'https://wiki.nixos.org/w/index.php?title=Special:UserLogin&returnto=NixOS+Wiki' | head -3
HTTP/2 429
server: Varnish
retry-after: 600
```

Reproduced from two unrelated client IPs (v4 and v6) on the first request each, so it is not a per-client penalty box.

This drops the two auth pages from the block list; the other URL classes are unchanged. The matching origin-side nginx rule was fixed and deployed in NixOS/nixos-wiki-infra#368.

cc @Mic92


Mic92:
applied.

Mic92:
Tested:

```
 - Special:UserLogin?returnto=… → 200
 - Special:CreateAccount?returnto=… → 200
 - action=edit → 429 (still blocked, as intended)
 - action=history&offset= → 429 (still blocked)
 - Main page → 200
```

PR #1212, opened 2026-09-14
builders: set LimitNOFILE on the queue-builder (+OOM avoidance)

vcunat:
Otherwise it retains soft limit of 1024, and that's probably causing issues sometimes.


builders/common/hydra-queue-builder.nix
```diff
@@ -29,4 +29,9 @@
       cpuPsiThreshold = 15.0;
     };
   };
+  # TODO: integrate this into hydra defaults?
+  systemd.services.hydra-queue-builder-dev.serviceConfig = {
+    ManagedOOMPreference = "avoid";
+    LimitNOFILE = 1024 * 1024;
```
Mic92, line 35: Sounds reasonable.

builders/common/hydra-queue-builder.nix
```diff
@@ -29,4 +29,9 @@
       cpuPsiThreshold = 15.0;
     };
   };
+  # TODO: integrate this into hydra defaults?
+  systemd.services.hydra-queue-builder-dev.serviceConfig = {
+    ManagedOOMPreference = "avoid";
```
Mic92, line 34: What is the effect of this?

vcunat: I copied it from older nixos infra's code for the queue-runner.  Details in `man systemd.resource-control.5`; TL;DR:
> only select this cgroup if there are no other viable candidates

vcunat:
There are nightly auto-deploys of the builders from the main branch, so this isn't deployed anywhere anymore.

builders/common/hydra-queue-builder.nix
Mic92: Okay, if you think this is still needed with queue runner v2?

vcunat: I don't know if it's needed.  It just seemed reasonable to still prefer killing e.g. processes of individual builds instead of the central thing (queue runner/builder).

vcunat: Or would the builds inherit this setting?  🤔

PR #1214, opened 2026-09-15
prometheus: scrape wiki node exporter via he1.wiki.nixos.org

Mic92:
wiki.nixos.org now points at Fastly, which doesn't expose :9100, so the node scrape job fails. Target the origin host directly.


vcunat (approved):
Looks OK, at a glance.

vcunat:
Deployed to `pluto`.

